PR #82, opened 2019-07-05
Feature/t5 our team

Napho:
Merge Our team page into develop branch to create base for other screens development


.env
```diff
@@ -0,0 +1 @@
+DATABASE_URL=postgres://di_website:chemical_ali!123!@db:5432/di_website
```
akmiller01: Probably should .gitignore the .env!

edwinmp: I remember deleting this file and adding it to .gitignore ... hmm

edwinmp: https://github.com/devinit/DIwebsite-redesign/commit/5948a5b1b797ccdec52103998cda91161f4605b0#diff-a084b794bc0759e7a6b77810e01874f2

di_website/ourteam/constants.py
```diff
@@ -0,0 +1,15 @@
+
+DEPARTMENTS =[
```
edwinmp: These are being added dynamically, so this isn't required anymore.

di_website/common/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
```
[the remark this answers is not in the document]
Napho: They are all helper classes and methods, i'd prefer to leave them there and just check first before re-writing any code that does the same thing, consider it a library with redundant helper methods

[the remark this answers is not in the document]
edwinmp: Exactly, redundant ... we may or may not use them. I'd rather you check in GNR, find what you need then add it to the project... it's what's proper. Otherwise we've got all this code or libraries whose purpose in the project is unclear... that's pollution.

.env
Napho: Removed unused code from common

di_website/common/apps.py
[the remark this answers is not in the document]
Napho: Removed unused code from common

PR #85, opened 2019-07-08
Feature/patterns

edwinmp:
More cherry picked changes from the pattern_library branch


akmiller01:
Worth correcting skipLinks again, @edwinmp ? Or are we expecting it just to change every time we pull in new code.
```
Error: Could not resolve './utils/skipLinks' from /home/alex/git/DIwebsite-redesign/patterns/assets/js/app.js
```

edwinmp:
> Worth correcting skipLinks again, @edwinmp ? Or are we expecting it just to change every time we pull in new code.
> 
> ```
> Error: Could not resolve './utils/skipLinks' from /home/alex/git/DIwebsite-redesign/patterns/assets/js/app.js
> ```

Ah, forgot to alert `fffunction` about that, and I couldn't find it in my code... do correct it.

PR #87, opened 2019-07-10
Fix/travis autodeploy

Napho:
Auto deployment to staging on develop branch


.gitignore
```diff
@@ -113,3 +113,4 @@ $RECYCLE.BIN/
 
 .env
 storage/
+deploy_key
```
edwinmp: New line ... motivation to start using VSCode... such stuff will be added automatically when u save the file hehe

akmiller01: Oh yes. Atom adds new lines at the end of files too, and has Python linting abilities.

PR #95, opened 2019-07-12
Changed implementation for UserProfile

akmiller01:


di_website/ourteam/models.py
```diff
@@ -65,6 +66,7 @@ class TeamMemberPage(Page):
         blank=True
     )
     telephone = models.CharField(max_length=255, null=True, blank=True)
+    my_story=RichTextField(blank=True,null=True,help_text="Please say something about team member ")
```
akmiller01: I was going to wait until @edwinmp merged in his paragraph streamfield to implement this @Napho. It would be nice to make all of the content fields a consistent streamfield.

PR #101, opened 2019-07-18
Fix/docker deploy

Napho:
Fix to have docker running against host directory  for django site


akmiller01:
Appears to successfully allow persistent files:
```
docker-compose exec web sh
/code # touch test
/code # exit
ls -lah test
-rw-r--r-- 1 root root 0 Jul 18 15:19 test
```

Issue #100, opened 2019-07-18
Changing the focal point on an image


k8hughes:
I tested the ability to change the image position in the hero for the our team page.  I wanted the position of this image to look the same way it does in the pattern library.  I tried a few different times to move the image focal point using the tool in the CMS but I don't think that any changes were coming through.

akmiller01:
This is being deployed right now @k8hughes . Do you want to give it a test and then close this issue if all is good? I've added closer cropping to the focalpoint.

Also, the image may not look the same as it does in the Pattern Library if you're missing hero-text. The height of that div is set by the amount of text, so quite a bit of the hero image is missing if you don't also have lorum ipsum in the hero text.

k8hughes:
LGTM - I've moved the focal point a few different ways and now it works perfectly.

akmiller01:
Brilliant. Now in future, if ever we want to make images more responsive to focal-point adjustments the way is `-c100`

PR #102, opened 2019-07-18
Fix/image issues

akmiller01:
In the event of a mis-loading image, prevents profile pictures from bleeding over:
image

Also added `-c100` to the hero so focal points will be more impactful. Might also ask Fffunction about setting min height? I think the majority of the issue is from the differing amounts of space according to the hero text.


akmiller01:
Should fix https://github.com/devinit/DIwebsite-redesign/issues/100

PR #103, opened 2019-07-19
Feature/blogs

akmiller01:
BlogIndexPage and BlogArticlePage

Tried to make `article_hero.html` reusable in case we need it for other types of articles, but we may want to move it/refactor it later. I'm also having trouble getting `BaseStreamBlock` to align content correctly @edwinmp . Some photos attached.

image

image

You can see the alignment issue with "Hello" on the left.


edwinmp:
@akmiller01 fixed the template & content issues above.

PR #104, opened 2019-07-23
Feature/patterns

edwinmp:


edwinmp:
@akmiller01 may as well approve and merge this in. It's stalling things. The correct image shall come in with the next PR.

PR #106, opened 2019-07-24
Fix/placeholder img

akmiller01:
Waiting on a gender-neutral placeholder. But hopefully I can just replace the .jpg when it comes in.


akmiller01:
Do we want to go ahead and merge, @edwinmp ? Or wait for the new placeholder?

edwinmp:
> Do we want to go ahead and merge, @edwinmp ? Or wait for the new placeholder?

@akmiller01  Let's wait for the new placeholder. Otherwise we'd have to do create a new PR for it as well.

di_website/settings/base.py
```diff
@@ -148,6 +148,7 @@
 
 STATICFILES_DIRS = [
     os.path.join(BASE_DIR, 'patterns/converted-html/assets'),
+    os.path.join(BASE_DIR, 'di_website/static'),
```
edwinmp: @akmiller01 I think this should be accessed through the patterns library and doesn't need for us to copy it into the `static` folder... will just need to merge in the `feature/patterns` once they have added the correct image, and you can access it in the template via `img/profile-fallback.jpg` ... if we already had an image with that name in place, we'd not have had to wait for the update.

akmiller01: Yeah we could do that. I was just worried that the pattern library would be overwritten and the image would disappear. But I guess it's in the pattern library now?

edwinmp: There's such worry for the whole thing, but it's unlikely for anything too destructive to get past us once we go live... not unless it's lost in a sea of changes.

Napho: would prefer to have all fffunction generated content in one place but either way i dont think place holders ever change so its still fine to have it in static

edwinmp: And what happens when we actually want a new placeholder from FF? Instead of just merging it in after they've made the change, we'd have to copy it again. We don't need that kind of duplication. We can keep the static folder, but not for stuff that's already in `converted_html`.

Issue #29, opened 2019-06-12
T11a News listing


k8hughes:
Sketch: https://marvelapp.com/3j51290/screen/57558637

Design: http://development-initiatives.surge.sh/page-templates/11a-news-listing

Functional Specification: 

Under the hero, we have a button to go to the newsletter sign up page. This is a page hosted on Mailchimp, so would be an external URL. This also applies to the footer button for the newsletter too. 

Below that the template lists the news posts. 

They should be ordered by date, so you see the latest posts first, and the ones further in the past last. 

There should be 9 posts shown on each page before the pagination controls. 

For each post, it should show: 

* Post title
* Post date
* Excerpt - this should be able to be set by the admin, and then fall back to the first 40 words in the content. Let's not use a character count to do this though - that could end up being a problem.

benjamincoleman:
> Change sign up button to be ‘sign up to our newsletter’ as that’s what people would be signing up to

edwinmp:
@Napho I'm not seeing much difference between this one and blog listing, so I'm banding them together.

k8hughes:
http://178.128.102.213/news/ 

I am assuming the empty white box is because the pages that will fill the box haven't been built yet.  I think this is working as intended.

edwinmp:
@k8hughes yes, that'll hold the "what we do" links.

Issue #21, opened 2019-06-12
T12a News single


k8hughes:
Sketch: https://marvelapp.com/3j51290/screen/57215884

Design: http://development-initiatives.surge.sh/page-templates/12a-news-single

Functional spec: 

Hero has the post title, post date and the label 'news'. 

Then there's richtext content. 

Then at the bottom we have related content - the approach to take with this is to allow admins to select up to 3 other blog posts, events, or publications to link to. If the admin selects none, or less than three, then fetch another blog post to add in here to make up the numbers to three. We'll have code in GNR to help with this is you want to get a head start.

edwinmp:
Hi @benjamincoleman ... you still have this in your design project, but had authorised and moved the news listing to the `Content Management` project. Should I move this one in too (if done)?

benjamincoleman:
@edwinmp I've been keeping cards in both projects, but we can make them only appear in one if that makes more sense project management wise.

edwinmp:
I don't mind them appearing in both. It's just that this one was still only in your project, so I assumed you still have some work to do on it.

benjamincoleman:
@edwinmp sorry about the late addition to this template - we need something added as a control or taxonomy to allow news posts to appear on the #32 page: 

> We then have a list of press releases, which are actually selected news posts. The news posts themselves would need a control to say "this post is a press release" and then allow it to be automatically shown in this template.

k8hughes:
There is some sort of error:  http://178.128.102.213/news/using-oda-blending-good-use-scarce-concessional-resources/

edwinmp:
Should be fixed in PR #115 ... @k8hughes  will let you know when merged.

edwinmp:
@k8hughes the fix has been deployed.

k8hughes:
Excellent now I can see the page and it all looks good.

PR #107, opened 2019-07-25
Feature/news

edwinmp:
- The What We Do sidebar is still pending, but all else should be done.

- [x] #21 
- [x] #29


akmiller01:
Is this section meant to show the tags? In the pattern library it just shows "NEWS"

image

image

edwinmp:
> Is this section meant to show the tags? In the pattern library it just shows "NEWS"
> 

I based this assumption on the old website, which had the tags on the article ... but now that I read the description, it's mean to just be 'News' ... or 'Blog' for blogs... hmm... alright.

benjamincoleman:
@edwinmp @akmiller01 the small labels are to indicate the type of content, not tags. We'll see them appear in other templates where we link to other pages that are related to the content of the current page. At that point it is helpful to indicate the type of content being linked to, whether that is a case study, publication, event, blog, news item etc. 

The old tagging system of the current website will no longer be used in this site.

edwinmp:
@akmiller01 added some changes... there's a requirement on the single blog & news pages for a limit of 3 on the related links, with the manually added links taking precedence. If they're less than 3, we're to fill in from the model.

di_website/templates/news/news_story_page.html
```diff
@@ -0,0 +1,42 @@
...
+<!-- TODO: include sidebar links from what we do once they're approved -->
+
+{% block after_content %}
+    {% if page.related_links %}
+        <section class="section">
+            <div class="row row--narrow">
+                <h2 class="section__heading">Related content</h2>
+                <div class="l-2up-3up">
+                    {% for related_page in related_pages %}
+                        {% with item=related_page.specific %}
```
akmiller01: Could you please add a {% if item.pk != self.pk %} check or something of the like? I've got one news page and it's linking to itself as a related page.

edwinmp: Ahh, great catch... thanks

di_website/news/models.py
```diff
@@ -0,0 +1,108 @@
...
+        context = super().get_context(request)
+
+        related_links = self.related_links.all()
+        related_links_count = len(related_links)
+
+        if related_links_count < MAX_RELATED_LINKS:
+            difference = MAX_RELATED_LINKS - related_links_count
+            news_pages = [link.page for link in related_links]
+            id_list = [page.id for page in news_pages]
+            news_objects = NewsStoryPage.objects
```
akmiller01: Ah just a moment. Also just caught that this should probably be `NewsStoryPage.objects.live()`

edwinmp: Done!

PR #122, opened 2019-08-01
Feature/breadcrumbs

edwinmp:
Had excluded breadcrumbs on vacancies ... this adds them. It's a template tag for reusability. Lots of pages coming up that'll need it.

### DIFF
--- a/di_website/home/templatetags/navigation_tags.py
+++ b/di_website/home/templatetags/navigation_tags.py
@@ -37,6 +37,7 @@ def get_menu_items(page, calling_page):
 
     return []
 
+
 @register.inclusion_tag('tags/navigation/primary.html', takes_context=True)
 def primary_menu(context, parent, calling_page=None):
     menu_items = get_menu_items(parent, calling_page)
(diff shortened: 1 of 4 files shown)


di_website/home/templatetags/navigation_tags.py
```diff
@@ -60,3 +62,17 @@ def secondary_menu(context, parent, calling_page=None):
...
+
+
+@register.inclusion_tag('tags/navigation/breadcrumbs.html', takes_context=True)
+def breadcrumbs(context):
+    self = context.get('self')
+    if self is None or self.depth <= 2:
+        # When on the home page, displaying breadcrumbs is irrelevant.
+        ancestors = ()
+    else:
+        ancestors = Page.objects.ancestor_of(self, inclusive=True).filter(depth__gt=1)
```
akmiller01: I don't have any intuition on how draft/live parent/child relationships work (can a child be live and a parent be draft?) Could you please change to `ancestors = Page.objects.live().ancestor_of(self, inclusive=True).filter(depth__gt=1)` just in case?

edwinmp: Ahh, sorry... I always forget this... thanks.

edwinmp: Done!

Issue #25, opened 2019-06-12
T6 Team member


k8hughes:
Sketch: https://marvelapp.com/3j51290/screen/57561812

Design: http://development-initiatives.surge.sh/page-templates/06-team-member

Functional Specification: 
This shows details of a team member. A team member can be featured on the team page #26, and they can also be used as an author for a blog post, or a publication, or for a case study. 

Model: 
* Portrait image
* Firstname - [text input]
* Lastname - [text input]
* Job title - [text input]
* Phone number - [text input]
* Email address - [email input]
* Biography - [content editor]
* Content by {firstname lastname} - this would be selected by fetching blog posts or publications where the person has been selected as an author.
* Other pages in this section - I would suggest you build this so an admin can add up to 4, and then the other spaces are filled by siblings of this page. You can see the siblings in the sidebar nav. All posts by a person will be shown (no pagination).

k8hughes:
This is working:  http://178.128.102.213/blog/new-world-bank-data-tell-us-progress-poverty/ 

You can see on this page Zach as the author

PR #144, opened 2019-08-07
Fix/related-pages

edwinmp:
- The other pages / related links sections where still visible even when there were no pages configured.
- The blog section for "Content by [User]" wasn't showing any content. #25

### DIFF
--- a/di_website/templates/ourteam/team_member_page.html
+++ b/di_website/templates/ourteam/team_member_page.html
@@ -1,7 +1,5 @@
 {% extends "layouts/default.html" %}
-
 {% load navigation_tags static wagtailcore_tags wagtailimages_tags card_tags team_tags static %}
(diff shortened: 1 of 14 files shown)


di_website/templates/ourteam/team_member_page.html
```diff
@@ -16,22 +14,29 @@
     <div class="row row--narrow">
         <div class="hero__profile">
             {% if self.image %}
-              {% image self.image fill-260x260-c100 as img %}
-              <img src="{{img.url}}" alt="{{self.name}} photograph">
+            {% image self.image fill-260x260-c100 as img %}
+            <img src="{{img.url}}" alt="{{self.name}} photograph">
```
akmiller01: Can we quickly add `width=260px height=260px` here to see if it lets us scale up smaller photos?

Issue #94, opened 2019-07-11
T8.1a - Place


benjamincoleman:
Created late on in the design process. 

Design - http://development-initiatives.surge.sh/page-templates/08-1a-place

Functional specification:

This template shows a single place.

We have the standard hero here, with page title and hero content, plus the image with credit if required.

There's the sidebar menu on the right, which has had a recent update to the way the title is spelled out "More about:" being the thing that has changed. What should happen there is the "More about:" will be fixed and the "What we do" will be the top page of the section that has this menu. So, What we do, Who we are etc. etc.

Under the hero we have a breadcrumb to aid wayfinding.

Under the breadcrumb we have standard editor content - text & images etc.

We have the ability to feature a country spotlight here. This modules consists of: 
- title
- content with formatting / links
- button with link
- image

Lastly, there is a "Related content" section. For this one that title should be editable, as we may want to have some fine grained control over that content. Then three items are selectable by the admin. These can pretty much be any type of content on the website.

edwinmp:
**NB:** We shall use a standard page to create this, so there won't be a specific "Place" page ... i.e once I'm done with "Where we work", the individual place pages shall be created based on the Standard Page template.

k8hughes:
And I have made a places page:  http://178.128.102.213/where-we-work/our-work-uganda/
I think there is some elements that are not the right way round
I expected the button to appear on the where we work page, next to a suggested place:  http://development-initiatives.surge.sh/page-templates/08a-where-we-work
instead it appears in the hero on the place page
this could be down to user error, but between these two templates I think you have some fields in the wrong place.

Napho:
> http://178.128.102.213/where-we-work/our-work-uganda/

Its a usability issue, the hero section used is generic for any standard page. Some standard pages have page links (like the one you created) while others don't. If you don't need a page link leave the **hero_link** & **hero_link_caption** fields empty from the admin section. It will be omitted

k8hughes:
Ok, this means the problem is with the parent page, I have mae a note on that issue.

edwinmp:
There's no problem ... you just have to use the correct stream fields. In this case, you have to use section stream fields. Also, creating a places page doesn't automatically add it to the "where we work" page... that's not how FF planned it to be. In the "where we work" page, you have to manually select the pages you want to link.

![image](https://user-images.githubusercontent.com/5672438/63160885-88081780-c027-11e9-9434-820995350deb.png)

Issue #93, opened 2019-07-11
T8a - Where we work


benjamincoleman:
Developed from T8 late in the process. 

Design: http://development-initiatives.surge.sh/page-templates/08a-where-we-work

Functional specification:

This template shows a single place. 

We have the standard hero here, with page title and hero content, plus the image with credit if required.

There's the sidebar menu on the right, which has had a recent update to the way the title is spelled out "More about:" being the thing that has changed. What should happen there is the "More about:" will be fixed and the "What we do" will be the top page of the section that has this menu. So, What we do, Who we are etc. etc.

Under the breadcrumb we have standard editor content - text & images etc.

Then we have a list of places we work. There are two types of this you can see in the template. We have Countries, and Regions. For both there are common elements and then just the one variation. Common elements are: 

- title
- subtitle
- content

Then we have the ability to link to "places" pages. These sit underneath this in the site heirarchy: 

![Screen Shot 2019-07-26 at 11 32 56](https://user-images.githubusercontent.com/64877/61945972-47256180-af99-11e9-84e2-5c26edc512a9.png)

As you can see from the design, the variation is the number of 'place pages' each 'place' links to. When it is a country, e.g. Nepal, it would have a button to link to a place page called Nepal. When it is a region, e.g. E Africa, it will have multiple buttons to cover a selection of places e.g. Uganda, Kenya. At the time of writing it is uncertain exactly how this content will be structured, so this will need to be flexible. I could be entirely possible that we could have a E Africa box, and only link to a page called E Africa. I'm happy to jump on a call to go through this, or preview things you're creating if that helps. 

~~Lastly, there is a "Related content" section. For this one that title should be editable, as we may want to have some fine grained control over that content. Then three items are selectable by the admin. These can pretty much be any type of content on the website.~~  - Removes from updated design

benjamincoleman:
Map - remove but use a screenshot as the hero image.

Places we have worked  ← Places we work

benjamincoleman:
Can we focus the screenshot on the equator please? 
Size for hero images = 1600px by 780px. Here's a PSD to help: https://www.dropbox.com/sh/hiy731zp3iyzt8a/AAC-Apn19-A1xEsTrYi80TnOa?dl=0

BenDarby:
Yeah sure, I don't have Photoshop so I'll do it in sketch

benjamincoleman:
[...]

PR #157, opened 2019-08-09
Feature/places

edwinmp:
- #94 
- #93 

Use `PlacesPage` to create the Where We Work page, and for each individual country or regional page, use the StandardPage.


di_website/templates/includes/partials/project.html
```diff
@@ -3,17 +3,17 @@
...
+      {% if self.image %}
+        {% image self.image fill-180x112-c100 as image %}
         <div class="project__media">
           <img src="{{ image.url }}" alt="{{ image.alt }}">
         </div>
       {% endif %}
-      <h2 class="project__heading">{{item.heading}}</h2>
-      <h3 class="project__subheading">{{item.sub_heading}}</h3>
+      <h2 class="project__heading">{{self.title}}</h2>
+      <h3 class="project__subheading">{{self.subtitle}}</h3>
```
akmiller01: Will we always have subtitles? If not, it may be worth piping this through `default_if_none:""` otherwise a null DB field prints the string `"None"`

edwinmp: Done!

Issue #39, opened 2019-06-12
T1b General with Accounts module


k8hughes:
Sketch: https://marvelapp.com/3j51290/screen/57215547

Design: http://development-initiatives.surge.sh/page-templates/01b-general-with-accounts-module

Functional Spec: 

This page is a variation of #19 

We have the standard hero here, with page title and hero content, plus the image with credit if required.

There's the sidebar menu on the right, which has had a recent update to the way the title is spelled out "More about:" being the thing that has changed. What should happen there is the "More about:" will be fixed and the "What we do" will be the top page of the section that has this menu. So, What we do, Who we are etc. etc.

Under the hero we have streamfield block from T1 #19 

Then we have a list of accounts groups. The design shows two groups, but there should be no limit to the number of groups. 

Each group has the following: 
- title
- introduction (paragraph + links)
- list of downloads. For each download we have
    - title
    - file downloads. Note this is the same content in different formats for accessibility reasons

---

This could just be text, and then it’s possible to 

a) grow over time  
b) be a nested tree map

I think that’s the best option given the anticipated low usage of this page.

---

@Darbice want to also include the filename for the download - so I've updated this in the sketch. This is true of anywhere we feature this download module.

![Screen_Shot_2019-05-29_at_15.40.51](https://gitlab.com/the-b-team/development-initiatives-website/uploads/f0b7f3e44cc3fb1eaaff648b91557b12/Screen_Shot_2019-05-29_at_15.40.51.png)

Review,Visual design,move-to-github

k8hughes:
I wanted to be able to add this page as it is described in the IA sitting as a child of "who we are" but the only child page allowed under  "who we are" is an "our story" page  which automatically incerpts a timeline.

@Napho what is hte name of the template in wagtail that I should be selecting?  Please can you allow it to be a child of "who we are"

edwinmp:
@k8hughes this should be supported under the existing General Page ... so I guess adding it as a child of all should solve this.

k8hughes:
http://178.128.102.213/who-we-are/our-funding2/ 

LGTM

alice-mcandrew-development-initatives:
@k8hughes I will test this when we have the content to upload for this page: http://159.65.56.142/who-we-are/our-finances/

alice-mcandrew-development-initatives:
[...]

Issue #19, opened 2019-06-12
T1: General Template


k8hughes:
Sketch: https://marvelapp.com/3j51290/screen/57215544

Design: http://development-initiatives.surge.sh/page-templates/01-general-page

Functional Specification:

This pages is used in several pages across the IA in: 

1. Our Strategy - Who we are > 
2. Consultancy work examples - What we do > Consultancy Services > 
3. Data support services - What we do > 
4. How we work, and sub pages of that page
5. Team stories - Work for us > 
6. Sub pages in the new Data Section (just out of design now)

**Hero**

- Image
- Image credit (capable of having a link in it)
- Title
- Subtitle

**Side navigation**

Shows pages in this section

Following the hero / sidebar nav we'd like a streamfield that can show the following items: 

- Content
- CTA module with quote
- Image and text module
- Video and text module
- Large image

I'll go through each in turn

**Content**

Under the hero we have a rich text editor content section

**CTA module with quote**

This same module is used on other templates such as #5 

- Image
- Image credit (capable of having a link in it)
- Module title
- Module content - rich text so can include links
- Module button - link location for the button
- Module button label - the text label of the button

**More content**

Following the CTA module, there's another section of rich text editor content

**Image / Video & Text modules**

Following the second set of content there's two examples of modules that have either a video, or an image, with accompanying text: 

<img width="1204" alt="Screen Shot 2019-08-27 at 15 44 30" src="https://user-images.githubusercontent.com/64877/63781419-a6043080-c8e1-11e9-9937-14c7cf28448e.png">

In each case there's the media item, and then on the right just a rich text block where the admins can enter headings, paragraph, links and formatting. 

**Large image with caption**

- Image
- Image credit (capable of having a link in it)

---

old content ⏬ 

# 🎯 Template Purpose

The general template is the base for all templates besides the Publications and Contact Us templates. 

# 🗂 Template details

**Template Number**

T1 General 

**Sketch:**

~~https://marvelapp.com/1b2f486g/screen/51364164~~

# 📋 Template spec

**Modules (as per spec for this template in IA):**

**Side menu**
* 4-6 Sub-menu items
* Active page - highlighted

**Page title**

*  Succinct one line overview or page title (See examples in [P1.0](~~https://marvelapp.com/1b2f486g/screen/51358448~~), P1.1, P1.2)

**WYSIWYG - sections with alternating background colour**

[...]

PR #180, opened 2019-08-20
Feature/general

Napho:
Merge general page (strategy & accounts) to develop
Refer to https://github.com/devinit/DIwebsite-redesign/issues/19 & https://github.com/devinit/DIwebsite-redesign/issues/39


di_website/templates/general/general_page.html
```diff
@@ -0,0 +1,42 @@
...
+        {% include_block content %}
+    {% endfor %}
+
+   
+    {% for block in page.downloads %}
+        {% include_block block %}
+    {% endfor %}
+    
+    <section class="section">
+        {% get_other_pages calling_page=self heading=page.other_pages_heading %}
```
akmiller01: Visually, this is rather similar to `blocks/banner` but I can see why you did it this way to avoid having to pick out team member photo, name, and position. Not sure if we want to refactor this just into a banner or not, but in the mean time, approve!

di_website/templates/general/general_page.html
```diff
@@ -0,0 +1,42 @@
...
+{% block hero %}
+    {% include 'includes/heroes/hero.html' %}
+{% endblock hero %}
+
+{% block body %}
+    <section class="section">
+        <div class="row row--narrow">
+            <div class="l-sidebar">
+                
+                {% include 'includes/partials/side_nav.html'%}
```
edwinmp: There's already a `secondary_menu` tag that handles this pretty nicely.

di_website/templates/general/general_page.html
```diff
@@ -0,0 +1,42 @@
...
+                        {% for block in page.body %}
+                            {% include_block block %}
+                        {% endfor %}
+                    </div>
+                </div>
+            </div>
+        </div>
+    </section>
+
+    {% include './includes/team_member_quote.html'%}
```
edwinmp: Ah, I was looking at code in the earlier files without knowing where it was going to be applied.
Anyway, I think all of these blocks can be added generically under one `SectionStreamBlock` ... allows for flexible ordering of the page content, including the banners and duos.

di_website/templates/general/general_page.html
```diff
@@ -0,0 +1,42 @@
...
+
+    {% for content in page.content %}
```
edwinmp: Right here... I handled this with `body` block for those under the side bar (typeset), and the `sections` block for those in the wide area of the page. i.e `TypesetBodyMixin` & `SectionBodyMixin`

di_website/templates/general/includes/team_member_quote.html
```diff
@@ -0,0 +1,32 @@
+{% load static wagtailcore_tags wagtailimages_tags string_utils %}
```
edwinmp: Think we can replace this entire file with a banner block

di_website/templates/includes/partials/side_nav.html
```diff
@@ -0,0 +1,22 @@
+<aside class="l-sidebar__aside">
```
edwinmp: Definitely shouldn't hardcode anything in this file ... the `secondary_menu` tag shall handle it... you just need to set it's parent to `self.get_parent`, and as long as these pages are tied to the correct parent (Who we are), they'll show the correct menu, even highlighting the current page.

di_website/common/blocks.py
```diff
@@ -194,6 +190,10 @@ class BaseStreamBlock(StreamBlock):
     link_block = LinkBlock()
     required = False
 
+class TypeSetImageBlock(ImageBlock):
```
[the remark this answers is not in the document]
Napho: I needed different templates for the image block and from what i see the only place you had used that image block was in that TypeSetStreamBlock, made sense to have it by name of section it where it was referred

di_website/common/blocks.py
```diff
@@ -194,6 +190,10 @@ class BaseStreamBlock(StreamBlock):
...
+    class Meta:
+        icon = 'fa-image'
+        template = 'blocks/image_block.html'
 
 class TypesetStreamBlock(StreamBlock):
```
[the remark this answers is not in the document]
Napho: Lemmi check that out

di_website/common/mixins.py
```diff
@@ -137,3 +137,34 @@ def get_title(self):
 
     def __str__(self):
         return self.title if self.title else self.file.title
+
+class TeamMemberStoryMixin(models.Model):
+
```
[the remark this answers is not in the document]
Napho: The banner requires that you pick everything from the go which i think is re-doing what is already done, again, a bit opinionated by id go with what you both think with @akmiller01

di_website/templates/includes/partials/side_nav.html
Napho: That came in the rebase, working on it

di_website/common/base.py
```diff
@@ -2,10 +2,8 @@
 from wagtail.snippets.models import register_snippet
 from wagtail.admin.edit_handlers import FieldPanel, MultiFieldPanel, PageChooserPanel
 from wagtail.images.edit_handlers import ImageChooserPanel
-
```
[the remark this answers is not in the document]
Napho: I don't think they are related

PR #181, opened 2019-08-21
Feature/services

akmiller01:
Probably all good, but I cannot test since the admin interface appears to be broken...
image

Thought you could use the `skill` section in common with whatwedo and services.


akmiller01:
Working now! CSS for Work Examples seem to be missing, however:

image

edwinmp:
The git history is broken ...

di_website/templates/whatwedo/services_page.html
```diff
@@ -0,0 +1,131 @@
...
+
+    <section class="section section--alt">
+        <div class="row row--narrow">
+            <h2 class="section__heading">Our skills</h2>
+            <div class="is-typeset max-meter">
+                <p>Our range of skills and expertise enables us to deliver our work:</p>
+            </div>
+            <div class="l-2up">
+                <div class="l-2up__col">
+                    <ul class="accordion">
```
edwinmp: Could you test with at least 4 skills and see if they'll appear in 2 columns...

di_website/templates/whatwedo/services_page.html
```diff
@@ -0,0 +1,150 @@
...
+                      </li>
+                      {% endif %}
+                      {% endfor %}
+                  </ul>
+                </div>
+            </div>
+        </div>
+    </section>
+
+{% if self.services_related_news %}
```
edwinmp: Just doing this doesn't work ... I usually go with `.count` i.e `self.services_related_news.count`

di_website/templates/whatwedo/services_page.html
```diff
@@ -0,0 +1,150 @@
...
+            <div class="l-2up-3up">
+              {% for item in self.services_related_news.all %}
+                  {% include 'includes/cards/card-news.html' with show_tag=True %}
+              {% endfor %}
+            </div>
+        </div>
+    </section>
+{% endif %}
+
+{% if self.services_related_example %}
```
edwinmp: Same here...

PR #184, opened 2019-08-26
Port over GNR content app into publications

akmiller01:


akmiller01:
@edwinmp Just rebased with develop and I'm still having an issue with the button class flipping correctly in the publication_page.html template. You can see the hard-to-read button at the bottom of this screenshot... Surge template doesn't seem to exhibit the same problem. Could you take a look?

image

akmiller01:
> @edwinmp Just rebased with develop and I'm still having an issue with the button class flipping correctly in the publication_page.html template. You can see the hard-to-read button at the bottom of this screenshot... Surge template doesn't seem to exhibit the same problem. Could you take a look?
> 
> image

Fixed it.

PR #189, opened 2019-08-30
Feature/error

davidebukali:


di_website/templates/404-static.html
```diff
@@ -0,0 +1,128 @@
...
+		<a href="#pagecontent" class="skiplink" data-js="skiplink">Skip to main content</a>
+		<header role="banner" class="header">
+            <div class="row">
+                <a href="/" class="branding">Development Initiatives</a>
+                <button role="button" class="navigation-primary-toggle" id="navigation-primary-toggle">
+                    <span>Menu</span>
+                </button>
+                <nav class="navigation-primary" role="navigation" id="navigation-primary">
+                    <ul class="navigation-primary__items">
+                        <li id="section-Who we are" class='navigation-primary__item'><a href="#">Who we are</a></li>
```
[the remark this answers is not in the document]
akmiller01: Ah yeah, forgot to mention @davidebukali that the static error pages need to be very bare-boned. They're rare to come up (NGINX rarely has a 500 error on its own), so you can do with a simple styled message with some hardcoded CSS.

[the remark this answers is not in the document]
davidebukali: @akmiller01 I used the link you shared at [Digital Ocean](https://www.digitalocean.com/community/tutorials/how-to-configure-nginx-to-use-custom-error-pages-on-ubuntu-14-04) to setup bare bones pages for nginx;

For wagtail errors i created a module called error;

To test set Debug = False and run server with --insecure flag because when Debug = False runserver does not handle static files and throws an error;

`ValueError: Missing staticfiles manifest entry for 'css/screen.min.css'  "GET /error-pagep/ HTTP/1.1" 500 59; `

Please see [this link](https://stackoverflow.com/questions/5836674/why-does-debug-false-setting-make-my-django-static-files-access-fail)

There is possibly a better solution to this using the original 

`STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'`

 but so far this is what I have found with 

`STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.StaticFilesStorage'`

di_website/error/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
akmiller01: Go ahead and remove this file since we're not using it.

di_website/error/models.py
```diff
@@ -0,0 +1,3 @@
+from django.db import models
```
akmiller01: I think you can delete this one too.

di_website/error/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
akmiller01: Delete this file.

di_website/error/views.py
```diff
@@ -0,0 +1,11 @@
...
+from django.http import HttpResponse
+
+
+def error_404_view(request, exception):
+    data = {"status": 404}
+    return render(request,'404.html', data)
+
+def error_500_view(request):
+    data = {"status": 500}
+    return render(request,'500.html', data)
```
akmiller01: Add blank line here. Unix likes a blank line at the end of files.

di_website/templates/404.html
```diff
@@ -1,9 +1,37 @@
...
+    </div>
+</section>
+
+<div class="section">
+    <div class="row row--narrow">
+        <div class="is-typeset as-wysiwyg max-meter">
+            <p>Sorry about that. Please use the navigation above to explore the website or contact us if you still can’t find what you’re looking for.</p>
+            <h4>Some things that might be helpful</h4>
+            <ul>
+                <li>Report this link as broken by emailing <a href="mailto:email@devinigt.org.uk">email@devinit.org.uk</a></li>
```
akmiller01: This should probably be info@devinit.org right?

di_website/templates/404-static.html
```diff
@@ -0,0 +1,47 @@
...
+                </div>
+            </section>
+
+            <div class="section">
+                <div class="row row--narrow">
+                    <div class="is-typeset as-wysiwyg max-meter">
+
+                        <h4>Some things that might be helpful</h4>
+                        <ul>
+                            <li>Report this link as broken by emailing <a href="mailto:email@devinigt.org.uk">email@devinit.org.uk</a></li>
```
akmiller01: info@devinit.org?

di_website/templates/404-static.html
```diff
@@ -0,0 +1,47 @@
...
+                            <li><a href="/" rel="home">Our homepage</a></li>
+                            <li><a href="#">Contact us</a></li>
```
akmiller01: Probably remove this link from the static.

di_website/templates/404.html
```diff
@@ -1,9 +1,37 @@
...
+                <li><a href="/" rel="home">Our homepage</a></li>
+                <li><a href="#">Contact us</a></li>
```
akmiller01: Try finding a way to dynamically link to the contact-us page. Maybe add a templatetag that looks for `ContactPage.objects.live().first()` and returns the page, and then check if that page exists, and if so  `{% pageurl contactpage %}`.

di_website/error/admin.py
davidebukali: Deleted admin.py

di_website/error/models.py
davidebukali: Also deleted models.py

di_website/error/tests.py
davidebukali: Deleted tests.py

di_website/error/views.py
davidebukali: Added blank line since we still need that `return render()` statement;

di_website/templates/404.html
davidebukali: I have updated, most definitely right; Second checked from the current contact us page [here](http://devinit.org/about/contact-us/)

di_website/templates/404-static.html
davidebukali: Updated

di_website/templates/404.html
davidebukali: Created template tag to load contact us link as you described.

di_website/templates/404-static.html
davidebukali: Done

di_website/templates/404.html
```diff
@@ -1,9 +1,39 @@
...
+
+<div class="section">
+    <div class="row row--narrow">
+        <div class="is-typeset as-wysiwyg max-meter">
+            <p>Sorry about that. Please use the navigation above to explore the website or contact us if you still can’t find what you’re looking for.</p>
+            <h4>Some things that might be helpful</h4>
+            <ul>
+                <li>Report this link as broken by emailing <a href="mailto:info@devinit.org">info@devinit.org</a></li>
+                <li><a href="/" rel="home">Our homepage</a></li>
+                <li>{% load_contactus %}</li>
```
akmiller01: Could you move the `<li></li>` into the template tag inside of the `if` check? Otherwise if there is no contactus_page, we get a blank bullet point.

davidebukali: Just moved the `<li>` and pushed

Issue #206, opened 2019-09-17
Simon’s request for something that will tell him when to check a page is up to date.


k8hughes:
There doesn't seem to be a plug in for this, so first step to starting the task is to have a conversation with Simon to fully understand the feature needs.  There is an email with some more detail that has been circulated but probably not enough detail.

edwinmp:
I added a Notification section to each page (the ones that were built at the time). You can write a message, specify the emails to send it to as well as the date/time. A cronjob on the server checks every hour and sends the emails that have reached their deadline.

I guess the only missing feature I can think of is recurrence, but so far, this should do. I ran it by Simon.

![image](https://user-images.githubusercontent.com/5672438/66553794-78eba480-eb54-11e9-8085-921ed01ee8c3.png)


![image](https://user-images.githubusercontent.com/5672438/66553815-83a63980-eb54-11e9-9506-f8c468312fc7.png)

k8hughes:
I've just set a notification, I'll check this off when I recieve the email.

k8hughes:
http://178.128.102.213/admin/pages/1310/edit/ 
att 11am I set an email to come at 12, but no email came. @edwinmp

k8hughes:
I got the emails in the end. LGTM.

alice-mcandrew-development-initatives:
I tried this and didn't receive the email, not sure why if it worked for Kate...

alice-mcandrew-development-initatives:
Simon is happy with that ^

edwinmp:
@Napho has the script and cronjob for this been setup on `pre-production`?

alice-mcandrew-development-initatives:
I just got an email notification...
![image](https://user-images.githubusercontent.com/55877063/69557279-3c59f800-0f9e-11ea-8138-baa20b74d7ef.png)
....which I was meant to get a while ago. This is what was set up in the cms:
![image](https://user-images.githubusercontent.com/55877063/69557325-4ed43180-0f9e-11ea-831c-9a6500d81720.png)

Is there a reason for this?
@k8hughes  I can set up another test?

k8hughes:
Yes, please set up another test and see what happens

alice-mcandrew-development-initatives:
Doesn't seem to be working

akmiller01:
First things first, the cronjob file wasn't even executable!
```
root@ubuntu-s-1vcpu-1gb-lon1-01:~# /home/di_website/send_page_notifications.sh
[...]

PR #225, opened 2019-09-24
Feature/page notifications

edwinmp:
Added a Notification model that can be linked to pages and used for configuring & scheduling notifications. Created a command that will then need to be called periodically to find and send the overdue notifications.

@akmiller01 could you help out with the script once deployed?

issue #206 

image


akmiller01:
I can help with the script, but we need to choose a level of involvement here with emails.

Easy way: Hook it up to a gmail account (I've got credentials for one called `devinitautomailer`) to send the emails.
Hard way: Run an independent email server.

edwinmp:
> I can help with the script, but we need to choose a level of involvement here with emails.
> 
> Easy way: Hook it up to a gmail account (I've got credentials for one called `devinitautomailer`) to send the emails.
> Hard way: Run an independent email server.

I used my Gmail account to test this, so I'm for that ... don't think we need anything fancy really, as long as it's a DI email. Love the automailer ... sounds perfect.

Issue #147, opened 2019-08-07
T21.2 Dataset single


benjamincoleman:
**Sketch:** https://marvelapp.com/3j51290/screen/59725631

There are three screens showing the contents of the tabs, so you can click on them in the Marvel link above and see them. 

**Design:** http://development-initiatives.surge.sh/page-templates/21-2-dataset-single

**Functional specification:** 

**Hero:**  

- Small tab with the Publication and the Figure in.
- Dataset title
- Optional description

Under the hero we have the Release date of the dataset

Then we have Wysiwyg content

There is an optional Alert to show if there is a more up to date version of this dataset, with a link to said dataset. 

Download module variant - note the explanation as to format types and what they are suitable for.

Analyst - would be DI team member, so can use the existing design pattern for authors. 

The content underneath the DI team member about contacting people will be standard text across all datasets

**Metadata tabs:** 

Will be accordion on mobile, tabs on large screens - let me know if you need JS help here. 

Tab 1 - Description
All Wysiwyg fields, but will be 5 separate blocks. The last 4 for provenance, variables, geography and topics will only show if they are completed. 

Tab 2 - Sources
Will be a list of links to other data sources - so a variant of the dataset design pattern. They have for each a name of the dataset, which is a combination of the organisation providing it and the specific dataset name. Then there's the access date, which is **not the same** as the released date. On some of these we have buttons to link to the dataset which could be in this website, or could be an external URL. 

Tab 3 - Licence and suggested citation
Text content for each - will be separate fields. 

**References to charts etc.** 

We use this section to link to where the dataset is used. 

Each shows the report image, date and title - this is what we list on the publications screen. 

Then for each publication it shows the list of links to figures that use this dataset.

**Related datasets**

Will link to other datasets in the system - so same design here as the listing page. 

---

It's worth looking at the microformats in designing and developing this page, as we would love this data to be discoverable / machine readable. https://developers.google.com/search/docs/data-types/dataset

benjamincoleman:
Have a content model which has been included in the IA and the Sketch. 

I've also developed a template to gather real content for the design to use: https://docs.google.com/document/d/1N7Pwn8j6cNjrBoubS-0OCBamV4BOYgxPccetoeMOu7o/

benjamincoleman:
@BenDarby looks lovely mate, really good. 

I have some edits: 

Alert alignment is somewhat off: 

[...]

Issue #146, opened 2019-08-07
T21.1 Dataset listing


benjamincoleman:
**Sketch:** https://marvelapp.com/3j51290/screen/59725629

**Design:** http://development-initiatives.surge.sh/page-templates/21-1-dataset-listing

**Functional specification:**

This page uses the fallback hero pattern, so there's no image, and the class `` is used to control the height. 

Then we have the sidebar menu, linking to other pages in this section.

Also some Rich Text Editor content.

There are 4 filters which filter the datasets to show relevant datasets. The filter relationship should be AND - we may want to change that to OR though, so please bear that in mind. 

Values for the filters are: 

**Report**

* Global Humanitarian Assistance
* Investments To End Poverty
* Global Nutrition Report
* Something else

**Data Source**

Latest thinking is: 

- UN OCHA - FTS
- OECD - CRS
- World Bank - PovCalNet
- World Bank - Data Bank
- World Bank - LSMS
- World Bank - CCKP
- UNICEF - MICS
- IMF - WEO
- N/A - DHS
- N/A - INFORM
- DFID - N/A
- UNDP - HDI
- N/A - IPUMS
- N/A - IDMC
- FCA - N/A
- IFRC
- ACAPS
- UNHCR

**Country**

Same as #15 

http://development-initiatives.surge.sh/page-templates/13-publications

**Topic** 

Latest thinking is: 

- Poverty
- Aid/ODA
- Humanitarian finance
- Domestic finance
- Development finance
- Transparency
- Data use
- About Development Initiatives

Then we have the list of datasets. For each dataset we are showing: 

- a small tab with the publication, and the figure this dataset relates to. 
- title
- release date

Each dataset links through to a single page about that dataset. 

Lastly we have the standard "More About" section that will have appeared on other pages in the site, so you should be able to extend it to here.

benjamincoleman:
@BenDarby can you make some updates for me? 

For each dataset in the listing, replace "updated" with "released".

In the filters, can you sort the first two filters for Report and Data Source alphabetically

The titles of the datasets are looooooong I'm afraid. For example: 

> ITEP 2018 Figure 3.1. Domestic resources in developing countries are 12 times larger than international finance flows to developing countries, of which commercial sources of finance dominate 

I'll get some more examples into a model for you.

benjamincoleman:
@BenDarby can you add the text "released: " infront of the date on the listing items please? 

> For each dataset in the listing, replace "updated" with "released".

[...]

PR #230, opened 2019-09-26
Feature/datasection

davidebukali:
Depends on [T21.2 Dataset Single](https://github.com/devinit/DIwebsite-redesign/issues/147) for a count of datasets in the website; Also on [T21.1 Dataset Listing](https://github.com/devinit/DIwebsite-redesign/issues/146) for a link to view all datasets.


di_website/datasection/admin.py
```diff
@@ -0,0 +1,3 @@
+from django.contrib import admin
```
akmiller01: Please delete this file. Whenever you start a new app in Wagtail, you generally don't need `admin.py` `tests.py` or `views.py`.

di_website/datasection/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
akmiller01: Please delete.

di_website/datasection/views.py
```diff
@@ -0,0 +1,3 @@
+from django.shortcuts import render
```
akmiller01: Please delete.

di_website/templates/datasection/data_section_page.html
```diff
@@ -0,0 +1,158 @@
...
+{% block hero %}
+  {% include 'includes/heroes/hero.html' %}
+{% endblock hero %}
+
+{% block body %}
+
+<section class="section">
+  <div class="row row--narrow">
+    <div class="l-sidebar">
+      {% comment %} TODO:: Add more about section here {% endcomment %}
```
akmiller01: Fun fact, you can also do single line comments like
`{# this is a comment #}`

di_website/datasection/models.py
```diff
@@ -0,0 +1,86 @@
...
+        )
+
+    ]
+
+    parent_page_types = ['home.HomePage']
+
+    class Meta:
+        verbose_name = "Data Section Page"
+
+    def getRandomQuote(self):
```
akmiller01: I think this will only ever return either the first or the second quote. Could you query the length of `self.quotes` and use that length to pick a random quote from the full list? Also, please make sure this function always returns something. As written, if the length of `self.quotes` is 0, the for-loop will never trigger and the function will never return.

di_website/datasection/admin.py
davidebukali: Just deleted all three.

di_website/datasection/tests.py
davidebukali: Just deleted

di_website/datasection/views.py
davidebukali: Also deleted

di_website/templates/datasection/data_section_page.html
davidebukali: This has cleaned up the comments section, thanks :)

di_website/datasection/models.py
davidebukali: Created a separate function to count number of quotes in a for loop; Also added a return statement at the end of `getRandomQuote()` and a check in template to hide html for when no quotes exist.

di_website/datasection/blocks.py
```diff
@@ -0,0 +1,54 @@
...
+
+class QuoteStreamBlock(StreamBlock):
+    team_member = TeamMemberQuoteBlock()
+    external_member = ExternalMemberQuoteBlock()
+    required = False
+
+
+class DataSupportServicesBlock(StructBlock):
+    title = TextBlock()
+    description = TextBlock()
```
akmiller01: Hi David. Had a chance to re-review. All of these look good except for most of them could you please add `required=False`? I think description, telephone, email, photographer, and photo credit all need `required=False`. Your template already has booleans for most of them so they don't appear if none, so users just need the ability to omit sections.

di_website/templates/datasection/data_section_page.html
```diff
@@ -0,0 +1,157 @@
...
+    <div class="l-4up">
+
+      {% for page_item in related_pages %}
+      <a href="#" class="card">
+        {% if page_item.specific.hero_image %}
+          {% image page_item.specific.hero_image max-400x250 as photo %}
+          <div class="card__media" style="background-image: url({{ photo.url }});">
+          </div>
+        {% else %}
+          <div class="card__media" style="background-image: url('img/fallback-image.jpg');">
```
akmiller01: Please use `load static` at the top and reference this fallback image as a static file. I'm getting a 404 when it wants to load.

di_website/datasection/blocks.py
davidebukali: Made all these fields optional

di_website/templates/datasection/data_section_page.html
davidebukali: Now using static tag and profile placeholder image as default when no image is provided..

PR #238, opened 2019-09-30
Feature/publication fixes

akmiller01:
- Adding Green as an option (may require CSS update?)
- Raw HTML field to avoid content loss on save
- Fixing various download bugs.
- Management command to swap content and raw content fields for legacy publications.


di_website/templates/publications/publication_page.html
```diff
@@ -74,4 +74,11 @@ <h2 class="type-m type-s--trailer">{{ page.download_report_title }}</h2>
         {% endif %}
     </div>
 </div>
+
+{% if page.publication_related_links|length > 0 %}
```
edwinmp: I wonder if this had the same requirement as the blogs, events & news ... if so, then it'd need more than this to load at least 3.

akmiller01: Some pages show related links. Some don't. So I assumed it was an optional thing.

akmiller01:
Need a re-request @edwinmp or @Napho cheers

Issue #236, opened 2019-09-27
Remove trailing comma from authors list on publications


edwinmp:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[x] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
Can we remove the comma at the end after the name of the last person listed under Author(s)

**Screenshots/GIF**
![Bug on Pubs pages](https://user-images.githubusercontent.com/5672438/65767744-71310680-e137-11e9-84c8-df50ab55cae8.PNG)

k8hughes:
@edwinmp is this a duplicate of #235 ?

edwinmp:
Nope ... same screenshot, different issue.

Issue #235, opened 2019-09-27
Display publication author title in singular tense when only one author


edwinmp:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[x] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the bug**
Can we automatically change Authors to Author if just one person is listed?

**To Reproduce**
Steps to reproduce the behaviour:
![Bug on Pubs pages](https://user-images.githubusercontent.com/5672438/65767648-1e574f00-e137-11e9-97ca-64e0c4a2112a.PNG)

edwinmp:
@akmiller01 any chance this is already fixed, or it's an FF thing?

akmiller01:
FF thing @edwinmp . The CSS turns the author UL into comma separated box.

edwinmp:
@benjamincoleman

benjamincoleman:
@BenDarby can we remove the comma from after every author please? 

```
.m-authors__list li:after {
    content: ",";
}
```

BenDarby:
@benjamincoleman updated markup and css to be like this now
```
<div class="m-authors">
    <h3 class="m-authors__title">Authors</h3>
    <p class="m-authors__copy"><a href="#">Emily Underhill</a>, Joe Bloggs, <a href="#">Amanda Hughes Reilley</a></p>
</div>
```

akmiller01:
https://github.com/devinit/DIwebsite-redesign/pull/273

alice-mcandrew-development-initatives:
Looks great @akmiller01

PR #273, opened 2019-10-16
Fix author markup

akmiller01:
#236 #235


alice-mcandrew-development-initatives:
@akmiller01 and @edwinmp, thanks for changing this.
Is it possible to delete the space between the name and the comma that follows? E.g. here the spacing looks slightly odd:
Capture

PR #340, opened 2019-10-30
Fix/the footnotes2

akmiller01:


akmiller01:
Also flagging this here before this PR is complete @BenDarby it seems like the `<sup>` tag has no effect on the text in a wysiwyg div:

image

PR #344, opened 2019-10-31
Fix summary label

akmiller01:


di_website/publications/models.py
```diff
@@ -418,7 +418,7 @@ class Meta:
 
     @cached_property
     def label(self):
-        return 'The summary'
+        return 'the executive summary'
```
edwinmp: all lower case?

akmiller01: Yeah, it gets put into "Read the executive summary"

Issue #330, opened 2019-10-28
Share chart link icon missing


alice-mcandrew-development-initatives:
## I'm submitting a ...
Bug
**Describe the Issue**
The link logo to the left of 'Share chart' on publication pages appears to be missing
**To Reproduce**
![Share chart](https://user-images.githubusercontent.com/55877063/67679152-5c3ad380-f980-11e9-8faf-50c9a5db5419.JPG)

akmiller01:
The link is there, it's just painted in a non-existent colour. Fixed it to match the page colour here: https://github.com/devinit/DIwebsite-redesign/pull/345/files

PR #345, opened 2019-10-31
Fix/share chart

akmiller01:


akmiller01:
https://github.com/devinit/DIwebsite-redesign/issues/330

Issue #265, opened 2019-10-15
Hero image caption on homepage


alice-mcandrew-development-initatives:
## I'm submitting a ...
Bug or question

**Describe the Issue**
There's nowhere to add a hero image caption/URL for the 'Homepage'

**To Reproduce**
1. Go to http://159.65.56.142/admin/pages/3/edit/

**Expected Behaviour**
On the 'Our story' page you can:
![E g  our story hero image caption](https://user-images.githubusercontent.com/55877063/66818929-fe48cd80-ef35-11e9-845e-b4ca5bfb03fc.JPG)

**Screenshots/GIF**
But you can't on the 'Homepage':
![No option to add hero credit on home page](https://user-images.githubusercontent.com/55877063/66818942-043eae80-ef36-11e9-86c4-bef3853889a2.JPG)

edwinmp:
@alice-mcandrew-development-initatives this was supported to be pulled from the featured publication itself, but that's broken at the moment.

alice-mcandrew-development-initatives:
Ok, I think I misunderstood. I don't think we've set a featured publication yet so can see if this works when we get to that!

alice-mcandrew-development-initatives:
Should we be able to click-through to the publication featured in the hero?

alice-mcandrew-development-initatives:
@edwinmp OK, I'll look out for when this is ready to review

PR #347, opened 2019-10-31
Fix hero_home button

edwinmp:
#265


k8hughes:
This is what I see as the button within the hero for the home page
Screenshot 2019-10-31 at 15 05 30

I do not think this is correct.

Issue #275, opened 2019-10-16
Sign up for vacancy alerts design


alice-mcandrew-development-initatives:
## I'm submitting a ...
Bug

**Describe the Issue**
Red line next to  'Sign-up line' is missing on Vacancies page.
Also, the last tick box is missing.
The functionality isn't working yet - presumably this will come later?

**To Reproduce**
http://159.65.56.142/work-for-us/vacancies/

![Sign up live](https://user-images.githubusercontent.com/55877063/66923112-290f5080-f020-11e9-948c-58fa3baddec9.JPG)

**Expected Behaviour**
http://development-initiatives.surge.sh/page-templates/18-vacancy-listing

![Sign up pattern](https://user-images.githubusercontent.com/55877063/66923118-2ad91400-f020-11e9-800b-b88f04a18bc1.JPG)

akmiller01:
@AnnaHope @k8hughes did we ever make a decision on whether this is going to be handled through the CRM or Mail Chimp?

akmiller01:
@AnnaHope if we're going with Mail Chimp, could you please:

1. Start a new mailing list for vacancy announcements.
2. Add a column to it for a sign-up to flag which departments they're interested in
3. Create a new web form in Mail Chimp with all of the tickboxes required above.

AnnaHope:
@akmiller01 ideally in the future it would be the CRM - the system isn't ready yet. Is it easy enough to switch over later though?

AnnaHope:
@jamesharle one on the list for you! Please do let @akmiller01 know when done so he can perform the necessaries :)

Anna

akmiller01:
Yes we could eventually switch the back-end of where the form points.

jamesharle:
Thanks both - @akmiller01 please find form set up to mirror pattern library here: http://eepurl.com/gIg_Hv

k8hughes:
@jamesharle I'm copying the question @akmiller01 asked you on slack in case you didn't see it:
```when you created the Mail Chimp form, did it give you a bit of code to embed?```

**Comment by akmiller01** [comment]
Here's the embed code @davidebukali You'll need to shim it into our form on the template, I don't think you'll need the embedded CSS or JS, but you may:
```
<!-- Begin Mailchimp Signup Form -->
<link href="//cdn-images.mailchimp.com/embedcode/classic-10_7.css" rel="stylesheet" type="text/css">
<style type="text/css">
[...]

PR #362, opened 2019-11-04
Feature/vacancy alerts design

davidebukali:
Fix for issue [275](https://github.com/devinit/DIwebsite-redesign/issues/275)


edwinmp:
@davidebukali there are some comments on issue #275 that may still be relevant to this. Could you check with @akmiller01 if they are.

akmiller01:
Ah yeah sorry @davidebukali I've got a web form from James that I need to connect to the back-end of this.

davidebukali:
@akmiller01 and @edwinmp thanks for the heads up; if there is a separate web form for this (to sign up for vacancy alerts), does the pattern design as well as this PR become obsolete.

akmiller01:
@davidebukali unfortunately I think it is going to need to go in a pretty different direction. We'll need to remove the snippets that store the email addresses internally, remove most of the POST work, and embed the Mail Chimp form in the template. The form itself is here https://devinit.us11.list-manage.com/subscribe?u=a829237ca0cf1470615c7f059&id=aee418056a But I'm waiting on James to post the embedable HTML so we can adapt it.

di_website/vacancies/models.py
```diff
@@ -97,33 +97,39 @@ def get_context(self, request):
...
-                            if first:
-                                form['alert_message'] = form.get('alert_message') + ' ' + department.name
-                                first = False
-                            else:
-                                form['alert_message'] = form.get('alert_message') + ', ' + department.name
-                    else:
-                        for department in all_departments:
-                            self.create_subscription(email.value, 'jobs', department)
-                        form['alert_message'] = ' all departments'
+                if gdpr:
```
akmiller01: I think this can even be simpler, @davidebukali when you changed the form action to the Mail Chimp website, the data is no longer POSTed back to the model, so there's no validation we can do on our end. It's all validated via Mail Chimp's JS.

davidebukali:
Yes @akmiller01 also removed the subscription code because form `POST` processing is handled by mailchimp

di_website/vacancies/models.py
davidebukali: The subscription code from users model seems unnecessary as well because it was being used in vacancies to subscribe users; Let me remove it

PR #364, opened 2019-11-05
Collectstatic as di_website

akmiller01:


akmiller01:
@edwinmp Done! Still need to scp it over to the respective severs before merging

Issue #360, opened 2019-11-04
Map on what we do page showing incorrect locations


AnnaHope:
At the mo, the map on the 'what we do page' is showing our offices but should ideally be completely separate for us to manually enter. Just to flag that some places we have done work don't appear anywhere on the website such as consultancy for someone else, so wouldn't be just where a publication is tagged as being in a particular country. http://159.65.56.142/what-we-do/

@edwinmp

edwinmp:
@AnnaHope a fix for this has been deployed.

AnnaHope:
@edwinmp the map has disappeared entirely? we can't figure it out from the cms...could you help? FYI @alice-mcandrew-development-initatives

edwinmp:
Hi @AnnaHope ... please add locations for the map to appear.

![image](https://user-images.githubusercontent.com/5672438/68940167-84fdfe00-07b3-11ea-81d6-b0ba27a028b4.png)

alice-mcandrew-development-initatives:
This was done :)

PR #370, opened 2019-11-06
Fix/add where we work locations

davidebukali:
Fix for [360](https://github.com/devinit/DIwebsite-redesign/issues/360)


di_website/whatwedo/blocks.py
```diff
@@ -81,3 +82,24 @@ class ExpertiseBlock(StructBlock):
...
     class Meta():
         template = 'blocks/expertise.html'
+
+
+class MapBlock(StructBlock):
+    """
+    Allows the addition of a single benefit
+    """
+    location = TextBlock()
+    address = TextBlock()
```
edwinmp: Should this be required?

di_website/whatwedo/blocks.py
```diff
@@ -81,3 +82,24 @@ class ExpertiseBlock(StructBlock):
 
     class Meta():
         template = 'blocks/expertise.html'
+
+
+class MapBlock(StructBlock):
```
edwinmp: Could you please rename this to LocationBlock

di_website/whatwedo/blocks.py
```diff
@@ -81,3 +82,24 @@ class ExpertiseBlock(StructBlock):
...
+    """
+    Allows the addition of a single benefit
+    """
+    location = TextBlock()
```
edwinmp: `name`

di_website/whatwedo/blocks.py
```diff
@@ -81,3 +82,24 @@ class ExpertiseBlock(StructBlock):
 
     class Meta():
         template = 'blocks/expertise.html'
+
+
+class MapBlock(StructBlock):
+    """
+    Allows the addition of a single benefit
```
edwinmp: `a single location`

di_website/whatwedo/blocks.py
```diff
@@ -81,3 +82,24 @@ class ExpertiseBlock(StructBlock):
...
+
+    class Meta():
+        icon = 'fa-map'
+
+
+class MapStreamBlock(StreamBlock):
+    """
+    Handles creation of where we work locations
+    """
+    add_location = MapBlock()
```
edwinmp: Just `location` is fine I think... can add any verbose naming for the UI in the meta class

di_website/whatwedo/blocks.py
davidebukali: `address = TextBlock()` should not be required, however `location = TextBlock()` is needed on the map to show the country name on mouse hover.

davidebukali: Just replaced `addresss` with an optional `mobile` field` which can be also be shown on the map.

di_website/whatwedo/blocks.py
davidebukali: Just done and pushed

di_website/whatwedo/blocks.py
davidebukali: Updated and pushed

di_website/whatwedo/blocks.py
davidebukali: Just updated

di_website/whatwedo/blocks.py
davidebukali: Also updated

PR #375, opened 2019-11-07
Fix/download block alt

edwinmp:


edwinmp:
> Does a field really require `required=False` if it has a default value?

I thought it doesn't ... sadly, for booleans, you can't leave that field unchecked ... it triggers the required validation error.

PR #388, opened 2019-11-13
Feature/chapter nav

akmiller01:


AnnaHope:
@akmiller01 @Napho @davidebukali @edwinmp  FYI short publications don't have chapters. The template for short publications was simply designed so that subheadings that denote sections within the publication can be seen and quickly jumped to. We also need to be able to share that section with a link, so people can be given a link that takes them to that specific section.

alice-mcandrew-development-initatives:
This is page I've created to test this: http://159.65.56.142/publication/preliminary-oda-data-2018-total-falls-for-second-consecutive-year/

akmiller01:
Yep @AnnaHope and @alice-mcandrew-development-initatives what I essentially had to do is hardcode an invisible "Chapter 1" for the Short publication pages inorder for the nav to appear for them. This should show up on staging shortly!

PR #469, opened 2019-11-18
Fix/seo2

akmiller01:


akmiller01:
@edwinmp as far as you know, the only two pages not inheriting from HeroMixin are HomePage and TeamMemberPage, right?

I had to make custom metadata mixins for pages without HeroMixin, and this code will break any page that doesn't have a custom mixin or a HeroMixin.

Issue #410, opened 2019-11-18
Column count for addresses on Who We Are page


benjamincoleman:
http://159.65.56.142/who-we-are/

We have 3 addresses at the moment, so the parent class `l-4up` should actually be `l-3up`.

Applies to this module: 

![Screen Shot 2019-11-18 at 11 47 19](https://user-images.githubusercontent.com/64877/69050094-3772d280-09f9-11ea-95e0-51cbe30cc445.png)

akmiller01:
How high do the column counts go @benjamincoleman ? Is 4 the maximum? We could write template tag that outputs the class as l-1up through l-4up depending on the length of the number of offices...

k8hughes:
LGTM.

PR #504, opened 2019-11-19
Add template tag to set column class based on number of office locations

davidebukali:
Fix for [#410](https://github.com/devinit/DIwebsite-redesign/issues/410)

### DIFF
--- a/di_website/about/templatetags/count_columns.py
+++ b/di_website/about/templatetags/count_columns.py
@@ -9,4 +9,4 @@ register = template.Library()
 def columns():
     number_of_offices = OfficeLocation.objects.exclude(
         latitude__isnull=True).exclude(longitude__isnull=True).all().count()
-    return number_of_offices
+    return max(2, min(number_of_offices, 5))
--- a/di_website/about/templatetags/count_columns.py
+++ b/di_website/about/templatetags/count_columns.py
@@ -1,0 +1,12 @@
+from django import template
+from di_website.vacancies.models import OfficeLocation
+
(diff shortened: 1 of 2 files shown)


di_website/about/templatetags/count_columns.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+register = template.Library()
+
+
+@register.simple_tag
+def columns():
+    number_of_offices = OfficeLocation.objects.exclude(
+        latitude__isnull=True).exclude(longitude__isnull=True).all().count()
+    return number_of_offices
```
akmiller01: Looks like a good fix @davidebukali but we might want to wait for feedback from @benjamincoleman on what classes are actually meaningful. I'm not sure whether `l-1up` would be valid, and I'm assuming we want to max out at `l-4up`

At which point this line could become: 
```
return max(2, min(number_of_offices, 4))
```
Which would bound it between 2 and 4.

davidebukali: Thanks @akmiller01; `return max(2, min(number_of_offices, 4))
` is a neat constraint

akmiller01: Just doing a quick search through the pattern library @davidebukali 

Just as I suspected, it starts at 2, but goes to 5. So you could use `max(2, min(number_of_offices, 5))`

davidebukali: @akmiller01 this was updated and pushed

PR #522, opened 2019-11-20
Fix/pub widgets

akmiller01:
Including section ID and footnotes


akmiller01:
Ah sorry @edwinmp pushed a few more tweaks. Rereview? :)

Issue #532, opened 2019-11-20
Dataset filters


urlsangel:
http://159.65.56.142/data/datasets/

Chrome/Mac

#### Labels vs values
See also this issue for similar issue with label vs value as first option: #497 

The filters initial values are labels rather than values as per the issue above, and are not consistent with that implementation (which uses a "Filter by..." convention).

A better implementation would be for the default value to be something like "All reports", "All data sources", "All topics", etc.

#### Label `for` values incorrect

The form `<label>` `for` attributes don't match the `id` of the associated `<select>` (they all use "team-filter"):

![Screenshot 2019-11-20 at 16 05 21](https://user-images.githubusercontent.com/560463/69255970-9c245d80-0bb0-11ea-843c-9f0b687fd1f8.png)

#### No values for `Topic` filter

There are no valid option values in the `Topic` filter - this shouldn't be displayed if there is no associated data available.

#### Query results
If more than one filter is selected - e.g. one for report and one for data source - in all cases tested no results were returned.

This is likely because the query is using AND (so only results that match both filters are returned).

If the query filter was amended to OR, then results would always be returned (assuming that valid filters are only ever displayed in the selects).

#### Empty results

Selecting Kenya as a geography returns no results:

http://159.65.56.142/data/datasets/?report=&source=&topic=&country=kenya

This can be mitigated by doing reverse lookups before presenting filter options, and only rendering options that have associated data.

**Note:** no results are returned when filtering on any geography, so it's possible that this filter doesn't function correctly.

k8hughes:
LGTM.
The default values are now "all" and the filters for topics have been tagged and are working.

PR #553, opened 2019-11-21
Fix/dataset filters

davidebukali:
Fix for issues in [#532](https://github.com/devinit/DIwebsite-redesign/issues/532)


di_website/datasection/migrations/0023_auto_20191121_0906.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 2.2.4 on 2019-11-21 09:06
```
akmiller01: Could you remove these all migrations @davidebukali ? I'm afraid they'll conflict with the current ones.

davidebukali: @akmiller01 deleted all conflicting migrations

PR #567, opened 2019-11-22
Feature/note block

akmiller01:


di_website/templates/blog/footnotes.html
```diff
@@ -0,0 +1,11 @@
+{% load string_utils wagtailcore_tags %}
+{% if page.footnotes_list %}
+    <div class="m-notes">
+      <h3 class="m-notes__title">Notes</h3>
+      <ul class="m-notes__list">
+        {% for item in page.footnotes_list %}
+          <li><sup><a id="{{ item.footnote_id }}" href="#{{ item.source_id }}">{{ forloop.counter }}</a></sup>{{ item.text | urlize }}</li>
```
edwinmp: I don't think it's supposed to have the `<sup>` element here; only in the text ... looks strange here

image

akmiller01: Copied the markup directly from Surge @edwinmp 
image

http://development-initiatives.surge.sh/page-templates/12-blog-single

edwinmp: Ah, in that case...

akmiller01:
Cheers for the review @edwinmp but I'm DEFINITELY going to need to merge migrations here, if not wipe out the old ones and redo...

akmiller01:
Ready for re-review @edwinmp

PR #569, opened 2019-11-22
Datasets and figures in publications

akmiller01:
image

### DIFF
--- a/di_website/publications/models.py
+++ b/di_website/publications/models.py
@@ -356,6 +356,16 @@ class PublicationPage(HeroMixin, PublishedDateMixin, ParentPageSearchMixin, UUID
         children += list(self.appendices)
         return filter(None, children)
 
+    @cached_property
+    def filtered_datasets(self):
+        results = []
+        all_pub_datasets = self.publication_datasets.all()
+        for pub_dataset in all_pub_datasets:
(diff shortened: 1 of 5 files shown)


di_website/publications/models.py
```diff
@@ -356,6 +356,16 @@ def meta_and_appendices(self):
         children += list(self.appendices)
         return filter(None, children)
 
+    @cached_property
+    def filtered_datasets(self):
+        results = []
+        all_pub_datasets = self.publication_datasets.all()
+        for pub_dataset in all_pub_datasets:
+            if type(pub_dataset.dataset.specific).__name__ == "DatasetPage":
```
akmiller01: This was ugly, but it's the only way to check type without a circular import.

di_website/publications/models.py
```diff
@@ -356,6 +356,16 @@ def meta_and_appendices(self):
         children += list(self.appendices)
         return filter(None, children)
 
+    @cached_property
+    def filtered_datasets(self):
```
edwinmp: This block of code is repeated 6 times ... could you extract it and reuse, cut down on the lines of code.

akmiller01: Made a mixin @edwinmp

Issue #525, opened 2019-11-20
Vacancies list could do with an empty list message


urlsangel:
http://159.65.56.142/work-for-us/vacancies/

Chrome/Mac

#### Vacancies list could do with an empty list message

(Unpublish the current test vacancy to replicate)

If there are no current vacancies in the list, then an alternative message would be helpful - currently the page has an empty space where the list would be:

![Screenshot 2019-11-20 at 14 43 01](https://user-images.githubusercontent.com/560463/69248629-7d1fce80-0ba4-11ea-905b-920cadc4a6dd.png)

k8hughes:
LGTM

PR #574, opened 2019-11-25
Show message for empty vacancy list

davidebukali:
Fix for [#525](https://github.com/devinit/DIwebsite-redesign/issues/525)

### DIFF
--- a/di_website/templates/vacancies/vacancy_index_page.html
+++ b/di_website/templates/vacancies/vacancy_index_page.html
@@ -47,12 +47,13 @@
           {% for vacancy in vacancies %}
             {% include 'includes/cards/card-job.html' %}
           {% endfor %}
-        {% else %}
-          <div class="is-typeset max-meter">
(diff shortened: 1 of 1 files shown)


di_website/templates/vacancies/vacancy_index_page.html
```diff
@@ -39,17 +39,22 @@
...
+    <div class="row row--narrow">
+      <div class="l-2up-3up">
+        {% if vacancies|length %}
           {% for vacancy in vacancies %}
             {% include 'includes/cards/card-job.html' %}
           {% endfor %}
-        </div>
+        {% else %}
+          <div class="is-typeset max-meter">
+            <p>No vacancies available</p>
```
akmiller01: We should ask comms for guidance on what you'd like for a no vacancy message. I would think something like "Sorry, we have no vacancy postings at the moment. Please subscribe to the mailing list below to be notified as soon as new vacancies are posted."

davidebukali: @akmiller01 comms provided guidance and have updated the text

Issue #579, opened 2019-11-25
Blog author


alice-mcandrew-development-initatives:
Firstly, some of the blog pages do not match the pattern library in terms of display of author:

E.g. http://159.65.56.142/blog/blog-blended-finance-impact/ (Should be on right not left)
E.g. http://159.65.56.142/blog/accelerating-crvs-and-legal-identity-in-africa-from-problems-and-principles-to-practice/ (Shown in middle instead of left, without image)

This is what it should look like: http://development-initiatives.surge.sh/page-templates/12-blog-single

Secondly, can I confirm that we have something in place to ensure that the hyperlink on the author's name is removed if there is no individual profile to link to (in the event that someone leaves)?

edwinmp:
@alice-mcandrew-development-initatives for the first link, the author is on the right on my laptop, which makes me believe it's one of those pesky screen size things.
As for the second one, this is the design for when a blog has multiple authors.

Since we're only going to unpublish the pages of past staff, we can check if an author's page is live before we create the hyperlink.

alice-mcandrew-development-initatives:
@edwinmp 
For the first, I think you might be right, as I'm just using laptop screen today (as I'm WFH) and it looks OK now!
For the second, did @AnnaHope sign off on this? I wasn't aware of discussions and there's no template to reference as far as I can tell.

So, if we unpublish a staff profile should the link styling automatically disappear then? (We wouldn't want it to look like a link and then go nowhere)

edwinmp:
Ah, @alice-mcandrew-development-initatives  I've figured out the difference... the author image appears on the left for blogs without a hero image
![image](https://user-images.githubusercontent.com/5672438/69619045-ffe4d580-104b-11ea-8a55-bcfcae464569.png)

alice-mcandrew-development-initatives:
@edwinmp That's good to know! We'll get heros for those pages!

edwinmp:
> For the second, did @AnnaHope sign off on this? I wasn't aware of discussions and there's no template to reference as far as I can tell.

@akmiller01 ^^

edwinmp:
> So, if we unpublish a staff profile should the link styling automatically disappear then? (We wouldn't want it to look like a link and then go nowhere)

@alice-mcandrew-development-initatives I'm currently implementing this ... will let you know when it's deployed.

akmiller01:
[...]

PR #584, opened 2019-11-26
Switch to div author template for unpublished or missing author pages

edwinmp:
Issue #579


akmiller01:
Oh also @edwinmp this include is also used on dataset pages.

edwinmp:
> Does it fix the padding issue? Even the ones with links still had the padding issue.

Don't think so ... basically just ensures we don't have links to unpublished author pages.

akmiller01:
Make sure you take a look at this locally for datasets to see if a div works in the layout.

edwinmp:
@akmiller01 tested and fixed.. could you please re-approve...

PR #588, opened 2019-11-26
Add inline panels to missing page models

akmiller01:


akmiller01:
> With all these crontab hustles, I consider using celery instead.

Maybe when we have time to refactor :)

Issue #519, opened 2019-11-20
No author meta on hero on guest blog article


urlsangel:
http://159.65.56.142/blog/future-humanitarian-transparency-assembling-big-picture/

Chrome/Mac

#### No author meta on hero on guest blog article

The [pattern library template](http://development-initiatives.surge.sh/page-templates/12a-blog-single-guest) has a guest author module in the hero:

![Screenshot 2019-11-20 at 11 38 15](https://user-images.githubusercontent.com/560463/69236318-2ce84280-0b8b-11ea-95cd-2b3d66aafac5.png)

The [dev implementation](http://159.65.56.142/blog/future-humanitarian-transparency-assembling-big-picture/) does not have this present:

![Screenshot 2019-11-20 at 11 38 06](https://user-images.githubusercontent.com/560463/69236339-3671aa80-0b8b-11ea-909a-24fa85837b02.png)

akmiller01:
In this particular case, there's no author so there's no link.

But in general, I think we need a CSS pull @edwinmp tried matching the markup exactly in https://github.com/devinit/DIwebsite-redesign/pull/523 (substituting an <a> for a <div> because it's reused in datasets), and it floats to the left instead of the right: 

Pattern library: 
![image](https://user-images.githubusercontent.com/2836840/69248487-61073780-0b7a-11ea-8032-ae2df9a00bc7.png)


Our page:
![image](https://user-images.githubusercontent.com/2836840/69248520-6ebcbd00-0b7a-11ea-972c-d60820bc1f1d.png)

k8hughes:
@alice-mcandrew-development-initatives I don't think we have any guest authors added to a blog yet.  Can you add one where it is appropriate to do so and then comment on the issue if there is still a problem or close it if you are happy?

AnnaHope:
@akmiller01 @k8hughes The implementation on guest blogs is still wrong. 

Should be the pattern library - http://development-initiatives.surge.sh/page-templates/12a-blog-single-guest

![image](https://user-images.githubusercontent.com/16958942/69555863-bd63c000-0f9b-11ea-813a-ffabb4ebc6c8.png)


Instead we have: http://159.65.56.142/blog/people-want-building-resilience-stability-conflict-crises/
![image](https://user-images.githubusercontent.com/16958942/69555949-dec4ac00-0f9b-11ea-8cec-ede5b7bb94d7.png)


Moving back to 'to do'

akmiller01:
[...]

PR #589, opened 2019-11-26
Update blog author image dimensions and alignment

davidebukali:
Fix for [#519](https://github.com/devinit/DIwebsite-redesign/issues/519)

### DIFF
--- a/di_website/templates/includes/heroes/hero.html
+++ b/di_website/templates/includes/heroes/hero.html
@@ -1,7 +1,7 @@
 {% load wagtailimages_tags wagtailcore_tags responsive %}
 
 {% responsiveimage page.hero_image fill-1600x780-c100 srcset=global.hero_srcs as image %}
-<section class="hero hero--image {% block hero_class %}{% endblock %}">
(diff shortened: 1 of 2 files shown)


di_website/templates/includes/heroes/hero.html
```diff
@@ -1,7 +1,7 @@
 {% load wagtailimages_tags wagtailcore_tags responsive %}
 
 {% responsiveimage page.hero_image fill-1600x780-c100 srcset=global.hero_srcs as image %}
-<section class="hero{% if image %} hero--image{% endif %} {% block hero_class %}{% endblock %}">
+<section class="hero hero--image {% block hero_class %}{% endblock %}">
```
edwinmp: Nope ... the `hero--image` class is appropriately conditional. This will break a lot of stuff that function recently complained about.

davidebukali: @edwinmp I noticed that `hero--image` floats the author meta to the right according to pattern template design; though as i have just read through [#579](https://github.com/devinit/DIwebsite-redesign/issues/579) you are definitely right.

PR #590, opened 2019-11-26
Fix/legacy footnotes and define no footnote richtext features

akmiller01:


akmiller01:
> I'm anticipating headaches arising from that name `RICHTEXT_FEATURES_NO_FOOTNOTES` ... you'd need to change it if this was excluding more than footnotes hehe

I'm hoping we don't have too many more changes to the richtext editors.

PR #595, opened 2019-11-27
Lift page restriction

akmiller01:
Awaiting new designs from Ben, but otherwise I think this is all we'd really need.


edwinmp:
@akmiller01 ah, sorry... think my #599 PR make this one obsolete ... should have just built on it, but just seeing it... oh well.

Issue #435, opened 2019-11-18
Static assets not using any hashing algorithm to enable caching


urlsangel:
http://159.65.56.142/

Mac/Chrome

#### Static assets not using any hashing algorithm to enable caching

Static assets such as `/assets/css/screen.min.css` and `/assets/js/app.bundle.js` are not being served with via [ManifestStaticFilesStorage](https://docs.djangoproject.com/en/2.2/ref/contrib/staticfiles/#manifeststaticfilesstorage) or a wrapper such as [WhiteNoise](http://whitenoise.evans.io/en/stable/).

Utilising such a static serving strategy would mean that long cache expiry times can be set on these assets for great performance on page revisits.

PR #597, opened 2019-11-28
Fix/static assets

davidebukali:
Fix for #435

Install dependency with `pip install -r requirements.txt`


akmiller01:
Okay. Finally sorted this one @davidebukali there's a bug in Wagtail where if you're using the Manifest you need to first collectstatic before any migrate.

PR #615, opened 2019-12-02
feature/spotlights-api

edwinmp:
Exposing page content for the spotlights pages ... we may need to create actual SpotlightPage pages and probably do this better, but for now, I'm going slow, not trying to think too far ahead until the designs come in... I need this for the `data-hub-spotlights` repo though.

### DIFF
--- a/di_website/api/admin.py
+++ b/di_website/api/admin.py
@@ -1,0 +1,3 @@
+from django.contrib import admin
+
+# Register your models here.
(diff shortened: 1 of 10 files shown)


edwinmp:
Ah, don't mind keeping it local for now though ... will close it till ready... wish there was a way of switching from active to draft.

edwinmp:
Now that we've gone live, I'd rather have this on the dev server ... I'd had to update the nginx settings to allow CORs for the assets folder. Not sure whether that's much of a security risk.

edwinmp:
@akmiller01 I need this on the dev server ... I don't see it having any effect on the live site.

Issue #605, opened 2019-12-02
Changes to the Vacancy page


riahasler-DI:
Can we add a Second Interview date in  the Dates section
Can we change the format of the Job start date to be just the year and the month (eg: January 2020) 

![image](https://user-images.githubusercontent.com/56585598/69953171-9a339600-14f0-11ea-9dd9-70b139eeecbc.png)

**Requested Delivery Date:** 02/12/2019

riahasler-DI:
I've also noticed that the optional fields still come through despite not adding content. 

![image](https://user-images.githubusercontent.com/56585598/70057267-c0326680-15d4-11ea-8d7c-3df1edf19a7d.png)

![image](https://user-images.githubusercontent.com/56585598/70057300-cfb1af80-15d4-11ea-8324-6f0b7c9c0f67.png)

riahasler-DI:
Can we make the Start Date a free text box so we have the option to put the full day month and year, the month and year and text (eg: Start Date: As soon as possible) 

Thanks! :)

PR #619, opened 2019-12-03
Fix/add vacancy page fields

davidebukali:
Fix for #605


di_website/templates/vacancies/vacancy_page.html
```diff
@@ -16,9 +16,10 @@
             <ul class="highlight__list">
                 <li><span>Location:</span>{{ page.location }}</li>
                 <li><span>Salary scale:</span>{{ page.salary_scale }}</li>
-                <li><span>Closing date:</span>{{ page.application_close_date|date:"j F Y" }}</li>
-                <li><span>First interview:</span>{{ page.first_interview_date|date:"j F Y" }}</li>
-                <li><span>Start date:</span>{{ page.job_start_date|date:"j F Y" }}</li>
+                <li><span>Closing date:</span>{{ page.application_close|date:"j F Y" }}</li>
+                <li><span>First interview:</span>{{ page.first_interview|date:"j F Y" }}</li>
+                <li><span>Second interview:</span>{{ page.second_interview|date:"j F Y" }}</li>
```
akmiller01: Should we wrap some of these in boolean if statements? I imagine not all of them will be filled out, and we don't necessarily want blank fields.

davidebukali: @akmiller01 great idea, just pushed an update

Issue #492, opened 2019-11-19
Empty paragraphs in rich text content


urlsangel:
http://159.65.56.142/what-we-do/news/development-initiatives-response-uk-voluntary-national-review/

Chrome/Mac

#### Empty paragraphs in rich text content

The main content has empty paragraphs, including one at the start:

![Screenshot 2019-11-19 at 14 54 22](https://user-images.githubusercontent.com/560463/69157273-a117dd00-0adc-11ea-8b65-903410d2a75c.png)

This means that the content does not get an increased-size first paragraph with a white box to the left, as per the pattern library template.

[Dev implementation](http://159.65.56.142/what-we-do/news/development-initiatives-response-uk-voluntary-national-review/):

![Screenshot 2019-11-19 at 14 55 40](https://user-images.githubusercontent.com/560463/69157462-d9b7b680-0adc-11ea-893b-c3c0548d14f9.png)

[Pattern library template](http://development-initiatives.surge.sh/page-templates/12a-news-single):

![Screenshot 2019-11-19 at 14 55 46](https://user-images.githubusercontent.com/560463/69157489-e9cf9600-0adc-11ea-8d80-a6a720cf87d2.png)

edwinmp:
@akmiller01 I think this is happening for imported content only.

urlsangel:
@k8hughes @AnnaHope @akmiller01 @edwinmp 

Just to clarify, this issue could be viewed as a content problem, but as empty paragraphs could always be entered by CMS editors in the future, and as they have a detrimental effect on the display (particularly the first paragraph style), you can also treat this as a development issue.

There are a couple of mitigations - either strip the empty paragraphs from the rich text content before saving to the database, or from the database content itself before rendering as HTML.

There are a couple of libraries that can help with a task like this:

- https://pypi.org/project/bleach/
- https://pypi.org/project/beautifulsoup4/

urlsangel:
This issue also applies to blog article pages, which means that the initial paragraph is not getting the correct style applied, e.g.:

http://159.65.56.142/blog/priorities-uks-incoming-secretary-state-alok-sharma/

alice-mcandrew-development-initatives:
I've amended in this instance: https://github.com/devinit/DIwebsite-redesign/issues/492
Now, it's just something to keep a note of and action wherever we notice?

urlsangel:
Hi @alice-mcandrew-development-initatives 

[...]

PR #620, opened 2019-12-03
Update rich text cleaner to replace <p>\s*</p>

edwinmp:
I hope this fixes issue #492 ... had trouble replicating the issue, but this is the big difference I found

### DIFF
--- a/di_website/common/templatetags/string_utils.py
+++ b/di_website/common/templatetags/string_utils.py
@@ -2,6 +2,7 @@ import os
 
 import bleach
 import uuid
+import re
 
 from django import template
 from django.utils import formats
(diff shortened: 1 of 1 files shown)


di_website/templates/blocks/paragraph_block_footnote.html
```diff
@@ -1,3 +1,3 @@
 {% load string_utils footnotes_utils %}
 
-{% content_with_footnotes self %}
+{% content_with_footnotes self|content %}
```
akmiller01: This probably won't make a difference. `content` is imported from string_utils and used in the `content_with_footnotes` function: https://github.com/devinit/DIwebsite-redesign/blob/develop/di_website/footnotes/templatetags/footnotes_utils.py#L4

edwinmp: Then I still have no clue why this http://178.128.102.213/blog/priorities-uks-incoming-secretary-state-alok-sharma/ isn't getting stripped ... let me compare the original content

di_website/common/templatetags/string_utils.py
```diff
@@ -71,14 +72,16 @@
...
         text,
         tags=WYSIWYG_LIST,
         attributes=ATTRS,
         styles=STYLES,
         strip=True,
         strip_comments=True
-    ).replace('<p> </p>', ''))
+    )
+
+    return mark_safe(re.sub(r'<p>\s*</p>', '', clean))
```
edwinmp: @akmiller01 not great at regular expressions, so not sure if this will do for all paragraph tags with any number of spaces... seems to be working, but... hmm

PR #664, opened 2020-01-31
Nest author in first loop block, rather than outside

akmiller01:
This will require some testing on dev first, but in theory it should end up looking like the pattern library.

### DIFF
--- a/di_website/templates/publications/publication_chapter_page.html
+++ b/di_website/templates/publications/publication_chapter_page.html
@@ -22,24 +22,14 @@
       {% include 'publications/chapter-nav/nav.html' with page=page.get_parent.specific menu_label='Contents' chapter_number=0 %}
     {% endif %}
 
-    {% if page.authors|length > 0 %}
-      <section class="section section--large">
-        <div class="chapter-block-wrapper">
-          <div class="row row--narrow">
-            <div class="larger-center">
-              <div class="chapter-block__item">
-                <div class="chapter-block_content">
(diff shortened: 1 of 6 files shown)


di_website/templates/publications/publication_chapter_page.html
```diff
@@ -22,21 +22,6 @@
       {% include 'publications/chapter-nav/nav.html' with page=page.get_parent.specific menu_label='Contents' chapter_number=0 %}
     {% endif %}
 
-    {% if page.authors|length > 0 %}
-      <section class="section section--large">
-        <div class="chapter-block-wrapper">
-          <div class="row row--narrow">
-            <div class="larger-center">
-              <div class="chapter-block__item">
-                <div class="chapter-block_content">
```
edwinmp: Strange, I thought I'd fixed this a long while back ... As far as I can tell, this class `chapter-block_content` is responsible for the difference with the PL. They should have updated this as well, or removed the styles completely. In fact, I think this is where the change should originate, not on our end ... the code below should be updated to match all else.
`
.chapter-block_content .m-authors {
    background-color: transparent;
    padding: 0;
    padding-left: 2em;
    border-left: 2px solid #ddd;
}
`

edwinmp:
@akmiller01 is there anything pending on this?

akmiller01:
I was waiting on approval of the appearance from @AnnaHope there are screenshots in issue 663.

AnnaHope:
@akmiller01 really sorry, didn't mean to hold this up! The screenshots look good to me so happy for you to make this live.

PR #665, opened 2020-02-06
Replace Google Analytics with Google Tag Manager

edwinmp:
Had made the previous PR to `develop`, but that's now saturated with Spotlight code, and this is needed live.


edwinmp:
> Question @edwinmp this doesn't remove analytics, just adding GTM right?

True ... I already configured GA to go through this as well. Actually, anything we configure through GTM should appear in our original analytics account.

PR #695, opened 2020-03-27
Feature/datasources text

davidebukali:
Add fields to the `SpotlightPage` model as well as the api to provide values for Data Sources Section in Spotlights


di_website/spotlight/models.py
```diff
@@ -18,12 +18,17 @@ class Meta():
     country_code = models.CharField(max_length=100, help_text='e.g. UG, KE', default='')
     country_name = models.CharField(max_length=255)
     currency_code = models.CharField(max_length=100, help_text='UGX, KES', default='')
-
+    datasources_description = models.TextField(help_text='A description for data sources section', default='')
+    datasources_url_title = models.TextField(help_text='A title for data sources url link', default='')
+    datasources_url = models.URLField(help_text='A url link to data sources for download', default='')
```
edwinmp: I'd rather this wasn't hardcoded ... the slug for the data sources page could change and we'd then have to come change this as well. Since it's an internal page, we'll have to select it.

di_website/spotlight/models.py
```diff
@@ -18,12 +18,17 @@ class Meta():
...
     content_panels = Page.content_panels + [
         MultiFieldPanel([
             FieldPanel('country_code'),
             FieldPanel('country_name'),
-            FieldPanel('currency_code')
+            FieldPanel('currency_code'),
+            FieldPanel('datasources_description'),
+            FieldPanel('datasources_url_title'),
+            FieldPanel('datasources_url')
```
edwinmp: Group these under a MultiFieldPanel.

di_website/spotlight/models.py
```diff
@@ -18,12 +18,17 @@ class Meta():
     country_code = models.CharField(max_length=100, help_text='e.g. UG, KE', default='')
     country_name = models.CharField(max_length=255)
     currency_code = models.CharField(max_length=100, help_text='UGX, KES', default='')
-
+    datasources_description = models.TextField(help_text='A description for data sources section', default='')
+    datasources_url_title = models.TextField(help_text='A title for data sources url link', default='')
```
edwinmp: Caption? ... title sounds like something above the actual URL :)

di_website/spotlight/models.py
edwinmp: Actually, instead of these static ones, could you please create a generic solution, kinda like how we do for related pages. That way we can add as many links as we want. Should allow addition of either a page or url & a caption. I think we have a streamfield with a similar structure. `AbstractLinkBlock`

davidebukali: Updated to use both a page chooser and url field

di_website/spotlight/models.py
davidebukali: Updated

di_website/spotlight/models.py
davidebukali: Yea, caption sounds about right, I updated this as well

di_website/spotlight/models.py
davidebukali: All datasource links are now added with a streamfield

di_website/api/utils.py
```diff
@@ -19,9 +21,17 @@ def serialise_page(request, page, fields=['title', 'full_url', 'active']):
...
-
     return result
 
+def serialiseDatasources(request, spotlightPage):
+    dataSourceLinks = []
+    for item in spotlightPage.datasources_links:
+        page = {}
+        page['caption'] = item.value['caption']
+        page['url'] = item.value['url']
+        page['page_url'] = item.value['page'].get_url(request, request.site) if item.value['page'] else ''
```
edwinmp: We need only one of these, so it's an either/or situation. Just `url` is fine, with the hardcoded url taking precedent. If there's no url, check for a page. Basically, don't want this `page_url` vs `url` check being done on the React side.

di_website/spotlight/models.py
```diff
@@ -3,11 +3,15 @@
 from wagtail.search import index
 from wagtail.core.models import Page
 from wagtail.core.fields import StreamField
-from wagtail.admin.edit_handlers import FieldPanel, MultiFieldPanel
+from wagtail.admin.edit_handlers import FieldPanel, MultiFieldPanel, StreamFieldPanel
 from wagtail.snippets.edit_handlers import SnippetChooserPanel
+from wagtail.core.blocks import (
+    StreamBlock
+)
```
edwinmp: One line for this, no braces. It's a single import.

di_website/common/blocks.py
```diff
@@ -110,6 +110,23 @@ class Meta:
...
+        help_text='For the link/button to show, either this or the url are required'
+    )
+    url = URLBlock(required=False, help_text='An alternative to an internal page')
+
+    class Meta:
+        icon = 'fa-link'
+
+class DatasourcesStreamBlock(StreamBlock):
+    link = DatasourcesLinkBlock()
+    required = False
```
edwinmp: If you're going to be using a StreamField, may as well reuse `LinkBlock`. This is an almost exact copy of it... The StreamField can be inlined in the model.

di_website/api/utils.py
```diff
@@ -19,9 +21,17 @@ def serialise_page(request, page, fields=['title', 'full_url', 'active']):
...
     result = object_to_dict(page, fields)
     result['relative_url'] = page.relative_url(request.site, request)
-
     return result
 
+def serialiseDatasources(request, spotlightPage):
+    dataSourceLinks = []
+    for item in spotlightPage.datasources_links:
+        page = {}
+        page['caption'] = item.value['caption']
```
edwinmp: The option to use the page title as a caption should be valid here (as I'm hoping u reused `LinkBlock`), in which case you'll have to handle that as well.

di_website/spotlight/models.py
```diff
@@ -18,13 +22,23 @@ class Meta():
...
     country_name = models.CharField(max_length=255)
     currency_code = models.CharField(max_length=100, help_text='UGX, KES', default='')
-
+    datasources_description = models.TextField(help_text='A description for data sources section', default='')
+    datasources_links = StreamField(
+        DatasourcesStreamBlock,
+        verbose_name="Datasources Links",
+        null=True,
+        blank=True
+    )
```
edwinmp: Won't need the StreamBlock
`datasource_links = StreamField([
        ('link', LinkBlock()),
    ], null=True, blank=True)`

di_website/spotlight/models.py
```diff
@@ -18,13 +22,23 @@ class Meta():
...
         MultiFieldPanel([
             FieldPanel('country_code'),
             FieldPanel('country_name'),
             FieldPanel('currency_code')
-        ], heading='Settings')
+        ], heading='Settings'),
+        MultiFieldPanel([
+            FieldPanel('datasources_description'),
+            StreamFieldPanel('datasources_links')
+        ], heading='Information for Datasources Section')
```
edwinmp: More than just information, right? `Data Sources Section` is fine.

di_website/api/utils.py
davidebukali: Updated this to be checked from the website

di_website/spotlight/models.py
davidebukali: Updated to one line

di_website/common/blocks.py
davidebukali: Now resusing `AbstractLinkBlock`

di_website/api/utils.py
davidebukali: Now uses the page url if caption is undefined

di_website/spotlight/models.py
davidebukali: Updated `datasource_links` to use this inline streamfield

di_website/spotlight/models.py
davidebukali: Yup, now reads just `Data Sources Section`

di_website/common/blocks.py
```diff
@@ -111,6 +111,11 @@ class Meta:
         abstract = True
 
 
+class DatasourcesStreamBlock(StreamBlock):
+    link = AbstractLinkBlock()
+    required = False
+
+
```
edwinmp, line 118: I meant you get rid of this.

di_website/common/blocks.py
```diff
@@ -111,6 +111,11 @@ class Meta:
...
 class LinkBlock(AbstractLinkBlock):
     class Meta:
         icon = 'fa-link'
```
edwinmp, line 121: In favour of this.

di_website/common/blocks.py
edwinmp: You're actually not using this at all.

davidebukali: Yes `DatasourcesStreamBlock` is unused, just removed it

di_website/common/blocks.py
davidebukali: Now using this inline

PR #739, opened 2020-05-05
Rearrange conf order to avoid rewrites and amend https server name

akmiller01:


edwinmp:
@akmiller01 is this still required?

akmiller01:
Yes still required @edwinmp Is https://data.devinit.org still redirecting for you? I'm afraid if we've overwritten django_https without this change it's just because it's a permanent redirect.

PR #744, opened 2020-05-08
Remove redundant title search index causing ES errors

akmiller01:


akmiller01:
Just a draft for now, as it's a fix in theory. I haven't been able to test yet.

Issue #618, opened 2019-12-03
Option to include infographics in blog post


alice-mcandrew-development-initatives:
If possible, it would be great to have the option to add infographics to blog posts (as we can for publications). 

Example: https://devinit.org/blog/how-well-aid-targeting-disability/
At the moment, we can add the charts as images but don't have the option for the title/caption styles.

edwinmp:
Hi @alice-mcandrew-development-initatives ... this is done. Please test on staging.

SimonMurphyDI:
Thanks @edwinmp this looks good to me

PR #745, opened 2020-05-09
Feature/blog infographics

davidebukali:
Fix for #618


di_website/blog/models.py
```diff
@@ -100,7 +117,10 @@ class BlogArticlePage(TypesetBodyFootnoteMixin, HeroMixin, Page):
         help_text='This date will be used for display and ordering',
     )
 
+    colour = models.CharField(max_length=256, choices=COLOUR_CHOICES, default=RED)
+
     content_panels = Page.content_panels + [
+        FieldPanel('colour'),
```
edwinmp, line 123: What's this supposed to do? As far as I can tell, all it does is change the colour of the infographic share button ... which doesn't make it very useful. It was for publications which had pages in all kinds of colours, but blogs are always in the classic red.

akmiller01:
> @akmiller01 could you please test this when you have a moment ... I'm not sure if it's behaving the same way for blogs as it does for publications.
> For example, clicking the share button show an empty input ... not sure if it's a config issue on my part, or some bug.
> image

Yeah we might want to code in a separate template for blog infographics without the share chart button and without the colour choices. The "Share chart" is dependent in the `uuid` feature of publications which blogs don't have.

akmiller01:
Yep, looking good! Just some conflicts to resolve with common/blocks @davidebukali  and then we can approve and merge
image

Issue #712, opened 2020-04-22
Creating a new child page (‘Audio and Visual Media Page’) for the ‘Publications’ parent page (now re-named ‘Resources’)


k8hughes:
**Is your feature request related to a problem? Please describe.**
Linked to the changes requested in #709 

**Describe the solution you'd like**
We want to be able to use a general page to deliver this, but to rename it within this section so it is easier for CMS users to know the right thing to select.

Please created a new template for the child pages of the ‘Publications’ parent page (which is now re-named ‘Resources’). This should be called: ‘Audio and Visual Media Page’. It should be based on the ‘General Page’ template with adaptions outlined below.
It should be imperative for the ‘Audio and Visual Media Page’ to be marked with the ‘Resource type’ snippet. Once marked, the ‘Resource type’ should appear as a label whenever the page is shown as a thumbnail, as well as at the top of each page, as outlined above.
Please allow for videos from Vimeo to be embedded on these pages, with the content appearing full-width on the page

Please ensure that you can add pages created using the ‘Legacy Publication’, ‘Publication Page’ and ‘Short Publication’ templates as ‘Related Links’ at the bottom of the page on any pages created using the ‘Audio and Visual Media Page’ template – and vice versa:›
![Screenshot 2020-04-22 at 16 59 41](https://user-images.githubusercontent.com/36768381/80004803-af6a8a80-84ba-11ea-99fe-35bcffd982f2.png)

**Describe alternatives you've considered**

**Additional context**

**Requested Delivery Date:** dd/MMM/YYYY

edwinmp:
@davidebukali the implementation for the two fields below is not flexible i.e. a user cannot reposition that content in different parts of the page as they desire. Please move the two fields below to the pre-existing stream fields i.e. TypesetBodyMixin & SectionBodyMixin ... There's already a video stream field in the TypesetBodyMixin ... the full width video should go in the SectionBodyMixin with all the other full width varieties.

![image](https://user-images.githubusercontent.com/5672438/81695886-e7d30800-946b-11ea-9605-a43a898785ab.png)

edwinmp:
But if the Audio & Visual Media Page will only have audio & video, then I suggest u create a separate StreamBlock for that and remove all the other mixins.

PR #749, opened 2020-05-13
Fix/flexible audio visual fields

davidebukali:
Fix for @edwinmp comment on issue #712


di_website/common/blocks.py
```diff
@@ -332,6 +333,17 @@ class Meta:
         icon = 'fa-image'
 
 
+class FullWidthVideoBlock(StructBlock):
+    full_width = EmbedBlock(
```
edwinmp: Why would you name it full width? Doesn't make sense... does it take in a full width?

di_website/publications/blocks.py
```diff
@@ -12,7 +12,7 @@ def render_basic(self, value, context=None):
 
         if value.type == 'video':
             player_code = '''
-            <h1 class="audio-player-title">Watch to video</h1>
+            <h1 class="audio-player-title">Watch video</h1>
             <div>
                 <video width="320" height="240" controls>
```
edwinmp, line 17: I haven't found any usage for this... it seems to only ever be audio

PR #775, opened 2020-06-11
Feature | visualisations

edwinmp:
Support for the addition of flexible visualisations to pages

@urlsangel added ace editor to help with JSON edits.


di_website/visualisation/blocks.py
```diff
@@ -0,0 +1,20 @@
+from django.utils.functional import cached_property
+from django.forms import CharField, Media
+
+from wagtail.core.blocks import FieldBlock
+from wagtail.admin.staticfiles import versioned_static
+
+from .widgets import AceEditorInput
+
+
+class ChartBlock(FieldBlock):
```
urlsangel, line 10: @edwinmp 

I'm not convinced of the benefit of using `Streamfields` as a data store unless there's really no other way (with regard to UI or nested lists for instance).

They're a performance hit, as they need to be JSONed in and out of the database, and add a lot of meta-info weight around simple data structures.

In this instance I would favour an `InlinePanel` data structure if requiring multiple JSON exports per page - which I'm not sure is necessary yet, we're still only using one JSON object per chart.

I would personally keep this simple and for now have a single field attached to the page instance, or refactor to an `InlinePanel` if you want to future proof with a repeatable structure.

A couple of other things:

- I'd use a `TextField` instead of a `CharField`, as that feels more appropriate to the data type
- I'd name the field something like `data` or `json` to make it more meaningful

di_website/visualisation/models.py
```diff
@@ -0,0 +1,51 @@
...
+    parent_page_types = ['home.HomePage']
+    subpage_types = ['visualisation.ChartPage']
+
+    class Meta:
+        verbose_name = 'Visualisations Page'
+
+    def get_sitemap_urls(self, request):
+        return []
+
+    def serve(self, request, *args, **kwargs):
```
urlsangel, line 24: @edwinmp 

As this is in essence a folder for the chart pages, I would say that the return value for the listing page would legitimately be a 404, as it should never be viewable:

```
from django.http import Http404

def serve(request):
    raise Http404()
```

di_website/visualisation/models.py
```diff
@@ -0,0 +1,51 @@
...
+from .blocks import ChartBlock
+
+
+class VisualisationsPage(Page):
+    """
+    Parent page for all visualisations
+    """
+    parent_page_types = ['home.HomePage']
+    subpage_types = ['visualisation.ChartPage']
+
```
urlsangel, line 17: @edwinmp 

As you only want one of these pages as a chart parent/folder, I would add this line to make the page unique:

```
max_count = 1
```

See: 
https://docs.wagtail.io/en/v2.9/reference/pages/model_reference.html#wagtail.core.models.Page.max_count

di_website/visualisation/models.py
```diff
@@ -0,0 +1,51 @@
...
+    def serve_preview(self, request, mode_name):
+        return self.get_parent().serve(request)
+
+
+class ChartPage(Page):
+    parent_page_types = [VisualisationsPage]
+    subpage_types = []
+
+    charts = StreamField([('chart', ChartBlock())], blank=True)
```
urlsangel, line 35: @edwinmp 

See note about using `InlinePanel` or regular field in `blocks.py`.

di_website/visualisation/models.py
```diff
@@ -0,0 +1,51 @@
...
+        StreamFieldPanel('charts')
+    ]
+
+    class Meta:
+        verbose_name = 'Chart Page'
+
+    def get_sitemap_urls(self, request):
+        return []
+
+    def serve(self, request, *args, **kwargs):
```
urlsangel, line 47: @edwinmp 

This method needs modifying to return a valid response if the user is authenticated.

If the user is authenticated (i.e. a logged-in CMS editor), it should return a simple template that displays the chart in isolation.

 I'm guessing this should be the same HTML include that is used to render the chart elsewhere on the site, a bare-bones HTML `<head>`, and whatever supporting JS is required.

We can attack this at some point in the pattern library if that helps?

If the user is not a CMS user, this method should return a 404 as per its parent page type.

di_website/visualisation/models.py
```diff
@@ -0,0 +1,51 @@
...
+        return redirect(self.get_parent().url)
+
+    def serve_preview(self, request, mode_name):
```
urlsangel, line 50: @edwinmp 

As previews are only available to logged-in CMS users, this should perform the same function as the `serve` method, returning a valid response that is the chart in isolation.

di_website/visualisation/static/visualisation/js/chart.js
```diff
@@ -0,0 +1,22 @@
+"use strict";
+const initPlotlyChart = () => {
```
urlsangel, line 2: @edwinmp 

Latest JS code is here, but is probably going to change as we progress - feel free to take a copy for now though: 
https://github.com/devinit/DIwebsite-redesign/blob/feature/visualisation_june_2020/patterns/assets/js/charts/entry.js

di_website/visualisation/static/visualisation/widgets/js/ace-editor.js
```diff
@@ -0,0 +1,40 @@
+"use strict";
+const renderChart = (options, element) => {
+    const { data, layout } = JSON.parse(options);
+    return Plotly.newPlot(element, data, layout);
+};
+const initPlotlyPreview = (widgetID, options) => {
```
urlsangel, line 6: @edwinmp 

This implies that editing of the exported JSON data can take place within the Wagtail page itself?

I'd be _very_ wary of allowing that, as you will likely end up with divergence, the danger being that an editor makes changes in Wagtail that are then overwritten by a new data export from Chart Studio.

di_website/visualisation/blocks.py
urlsangel: Hi @edwinmp 

I remembered the Postgres-specific JSONField after adding this comment, this would be even better suited to the data type:

https://docs.djangoproject.com/en/3.0/ref/contrib/postgres/fields/#jsonfield

edwinmp: Nice, I didn't know about this one ... will switch to this.

di_website/templates/publications/publication_chapter_page.html
```diff
@@ -82,8 +82,17 @@ <h2 class="type-m type-s--trailer">{{ page.download_report_title }}</h2>
...
+  {% with heading='Related content'  %}
+    {% include 'includes/partials/related_content.html' %}
+  {% endwith %}
 {% endif %}
+
+
+{% block html_footer %}
+  {{block.super}}
+
+  <script src="https://cdn.plot.ly/plotly-basic-latest.min.js"></script>
```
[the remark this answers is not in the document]
edwinmp: Changed to a Webpack implementation with dynamic imports ... Going to use Plotly [modules](https://github.com/plotly/plotly.js#modules) to import smaller files based on the chart type

di_website/visualisation/static/visualisation/widgets/js/ace-editor.js
edwinmp: I believe allowing editing within Wagtail is important since we don't have a finite number of charts on Plotly Studio under the free accounts we intend to use. That means as more and more charts are added, we'll depend more on Wagtail if we want to edit the older charts. Also, it makes things easier for advanced users.
To that end, I'm using a code formatter to make Wagtail edits easy:
image

urlsangel: @k8hughes @benjamincoleman @edwinmp 

I understand that this feature may be useful, but my understanding of the process was that Wagtail was to be used as a data store for chart data, as opposed to an editing interface.

Here's a scenario in which I see a risk:

- User A creates a chart in Chart Studio, exports the data, and adds the data to a Wagtail chart page
- User B modifies the data in the Wagtail page using the JSON editor
- User A refines the chart in Chart Studio, re-exports the data, and updates the existing Wagtail page

In this instance, the modifications made by User B are lost as the chart no longer has a single source of truth (Chart Studio).

Just thought I'd flag this up here as this feels like this implementation is diverging a little from the process that's been discussed so far.

PR #782, opened 2020-06-15
User selected sort order

akmiller01:


akmiller01:
Hm, doesn't seem to be us that have ruined the mobile view @edwinmp , there's a hard-coded 19% width on that class...
image

Also the selection does persist, but filtering by relevance only persists when "Search..." is filled out because it's not valid without a search query.

akmiller01:
Pushed a CSS fix @edwinmp 

Mobile
image

Desktop
image

Issue #774, opened 2020-06-11
Changes to the foreword in reports


k8hughes:
The current website functionality does not allow for any content to appear before the executive summary module however we'd like to have a forward first  – I know that GNR website was updated so foreword was optional so may be fairly straightforward if we have that code already.

@urlsangel could you point us in the right direction so we can grab this code from GNR?

urlsangel:
Hi @k8hughes @edwinmp 

The foreword was added in this commit:
https://github.com/devinit/gnr-website/commit/8857ff106691c51aa2422aa24bea55bd11a24bd9#diff-44bff827b72d070a8094405a29be021a

It was then also added to the home page listing here (not sure if that is relevant for the DI site though):
https://github.com/devinit/gnr-website/commit/714105ab11ccc1683dd2197264780480173f1945#diff-44bff827b72d070a8094405a29be021a

The `ReportForewordPage` type is essentially a clone of the `ReportSummaryPage` type - there are a number of places in the code where an optional foreword is added to the listing etc. if one is created and is 'live' (as with the summary page type).

It's probably worth Edwin reviewing the `reports/models.py` and associated files in the first commit referenced above to see where these changes need to be implemented in the DI site.

This task may need a bit of back and forth, so please feel free to give me a shout if you need to.

k8hughes:
@davidebukali This could be a good one for you to look into

benjamincoleman:
@k8hughes @edwinmp @davidebukali Also, let me know if you want the pattern library updated.

k8hughes:
@alice-mcandrew-development-initatives I'm passing this one over to you to test as I'm not sure where to go in the CMS to edit reports, can you check that the forward is working as intended?

alice-mcandrew-development-initatives:
@k8hughes and @davidebukali 
I've tested here: http://staging.devinit.org/resources/global-humanitarian-assistance-report-2019/

1) Please can we have the buttons on the parent page as lower-case for consistency with other modules:
![image](https://user-images.githubusercontent.com/55877063/85423351-82942b80-b56e-11ea-841d-b038b76e0586.png)

2) Please can we have the option to have the foreword in different colour. It looks like this on other child pages: 
![image](https://user-images.githubusercontent.com/55877063/85423913-2ed61200-b56f-11ea-9b51-d61822514ee9.png)

[...]

PR #807, opened 2020-06-26
Update the foreword section

davidebukali:
Fixes for comments [here](https://github.com/devinit/DIwebsite-redesign/issues/774#issuecomment-648244816)


di_website/publications/models.py
```diff
@@ -574,6 +577,10 @@ def label_type(self):
     def label(self):
         return 'the executive summary'
```
edwinmp, line 578: I think the comment wanted this as `executive summary` ... please verify.

di_website/templates/downloads/download-list.html
```diff
@@ -19,34 +19,35 @@ <h2 class="type-m type-s--trailer">{{ page.publication_downloads_title }}</h2>
                 {% endfor %}
             </ul>
         {% endif %}
-        {% if page.data_downloads_list or page.filtered_datasets %}
```
edwinmp, line 22: hmm ... by removing this check, the title on line 23 and any other conditional markup will always be shown even when there's no downloads ... 
This is the result:
image

And since this is a reused template, this affects about 6 types of pages. Dejavu!

di_website/templates/publications/chapter-nav/nav-item.html
```diff
@@ -18,7 +18,15 @@
...
     >
-      <span class="chapter-title">{{ item.label | title }}:</span>
+      <span class="chapter-title">
+          {% if item.is_foreword %}
+              {{ item.label }}:
+          {% elif item.is_summary %}
+              {{ item.label_type }}:
+          {% else %}
+              {{ item.label }}:
+          {% endif %}
```
edwinmp, line 28: I don't see where this change has been requested...
For summary pages, this changes the chapter title to `summary`, and that's not what Alice wanted, unless I'm missing something.
https://github.com/devinit/DIwebsite-redesign/issues/774#issuecomment-648244816
Side note, the logic would be the same even without checking for `item.is_foreword`.

di_website/templates/publications/partials/downloads.html
```diff
@@ -11,7 +11,13 @@
             <div class="orderReport__img" style="background-image: url({{cover_image.url}});"></div>
           {% endif %}
           <div class="orderReport__content">
-            <h2 class="type-m type-s--trailer">{{ page.download_report_title }}</h2>
+            <h2 class="type-m type-s--trailer">
+              {% if publication_downloads_title %}
```
edwinmp, line 15: This won't ever be true ... I think you meant `page.publication_downloads_title`. Which makes me wonder whether you tested this. And if u tested it and found it fine, then this change was unnecessary.
Why did you changes this?? BTW, I'm using the comment you tagged in this PR as a reference for your changes.

If this is markup for the section in the screenshot below, then I think it was fine as it was.

di_website/templates/publications/publication_page.html
```diff
@@ -31,12 +31,28 @@
...
+                            {% else %}
+                                {{ item.label }}
+                            {% endif %}">
+                          <span>
+                            {% if item.is_foreword %}
+                                {{ item.button_titles }}
+                            {% else %}
+                                {{ item.label }}
+                            {% endif %}
+                          </span></a>
```
edwinmp, line 45: Finding it hard to make sense of this ... probably needs some indenting, though I think it'll be better on one line, however long.

di_website/templates/publications/publication_foreword_page.html
```diff
@@ -12,15 +12,19 @@
...
-  <section class="section {% if forloop.first %}section--post{% endif %}">
-    <div class="row row--narrow">
-      {% if block.block_type == 'rich_text' and forloop.first %}
-        {% include_block block with id=block.id|string_start classname='is-typeset--additionals' %}
-      {% else %}
-        {% include_block block with id=block.id|string_start %}
-      {% endif %}
-    </div>
-  </section>
+  {% for block in page.content %}
```
edwinmp, line 15: Ah, this was my bad I think ...

di_website/templates/publications/publication_page.html
davidebukali: Updated cached property to use the `item.label`

di_website/templates/publications/chapter-nav/nav-item.html
```diff
@@ -18,8 +18,20 @@
...
     >
-      <span class="chapter-title">{{ item.label | title }}:</span>
-      <span class="chapter-nav__title">{{ item.title }}</span>
+      <span class="chapter-title">
+          {% if item.is_foreword %}
+              {{ item.label_type | capfirst }}
+          {% else %}
+              {{ item.label }}
+          {% endif %}
+      </span>
```
edwinmp, line 27: I don't see why you're changing this ... if you just set the label if the `PublicationForewordPage` to `foreword` instead of `the foreword`, this would stay as it was.

di_website/templates/publications/chapter-nav/nav-item.html
```diff
@@ -18,8 +18,20 @@
...
+      <span class="chapter-nav__title">
+          {% if item.is_summary %}
+              {{ item.label | capfirst }}
+          {% else %}
+              {{ item.title }}
+          {% endif %}
+      </span>
```
edwinmp, line 34: I don't see the reason for this change either ... This is meant for the title ... no one's said that Summary pages should only have titles `Executive summary` ... the label tells you it's a summary ... title can be anything.

di_website/publications/models.py
```diff
@@ -492,27 +495,27 @@ class Meta:
 
     @cached_property
     def label(self):
-        return 'The Foreword'
-
-    @cached_property
-    def is_foreword(self):
-        return True
+        return 'the foreword'
```
edwinmp, line 498: Thought this was to be changed to `foreword`

Issue #802, opened 2020-06-23
Setup a zero downtime deployment


mariamiah:
**Is your feature request related to a problem? Please describe.**
Currently there is downtime whenever a fresh deployment is made which makes the application inaccessible for a few minutes.. 

**Describe the solution you'd like**
There is need to set up a zero downtime deployment

**Describe alternatives you've considered**
Blue green deployment  where there is gradual transfer of user traffic from a previous version of an app  to another version—both of which are running in production. 

The old version may be the blue environment while the new version may be the green environment. Once production traffic is fully transferred from blue to green, blue can be stopped

**Additional context**

**Requested Delivery Date:** dd/MMM/YYYY

edwinmp:
@mariamiah anything pending on this? If not, move to `Deploy to live`

PR #808, opened 2020-06-26
Fix/deployment

mariamiah:
#802


.travis.yml
```diff
@@ -11,18 +11,14 @@ before_install:
...
-   on:
-     branch: master
+deploy:
+  - provider: script
+    skip_cleanup: true
+    script: cd .. && chmod 600 ./deploy_key && scp -o StrictHostKeyChecking=no -i ./deploy_key -r ./deploy.sh root@68.183.47.234:~/ && ssh -o StrictHostKeyChecking=no -o SendEnv=BRANCH -o SendEnv=CELERY_BROKER_URL -o SendEnv=DEFAULT_FROM_EMAIL -o SendEnv=EMAIL_BACKEND -o SendEnv=EMAIL_HOST -o SendEnv=EMAIL_HOST_USER -o SendEnv=EMAIL_HOST_PASSWORD -o SendEnv=DATABASE_URL -o SendEnv=ELASTIC_PASSWORD -o SendEnv=ELASTIC_SEARCH_URL -o SendEnv=ELASTIC_USERNAME -o SendEnv=HS_API_KEY -o SendEnv=HS_TICKET_PIPELINE -o SendEnv=HS_TICKET_PIPELINE_STAGE -o SendEnv=RABBITMQ_PASSWORD -o SendEnv=ENVIRONMENT -o SendEnv=SECRET_KEY -i ./deploy_key root@68.183.47.234 ./deploy.sh run
+    on:
+      branch: fix/deployment
```
edwinmp, line 24: won't this need to change?

config/nginx/django.conf.ctmpl
```diff
@@ -88,7 +120,6 @@ server {
 
     rewrite ^(/publications)(.*)$ /resources$2 permanent;
 }
-include /etc/nginx/conf.d/django_https;
```
edwinmp, line 91: Shall we still be able to run on https without this?

deploy.sh
```diff
@@ -153,16 +168,23 @@ function start_link_checker_processes {
...
+        log "user already exists. Skipping ..."
+        docker-compose exec -T rabbitmq rabbitmqctl add_vhost myvhost
+        docker-compose exec -T rabbitmq rabbitmqctl set_user_tags root root
+        docker-compose exec -T rabbitmq rabbitmqctl set_permissions -p myvhost root ".*" ".*" ".*"
+    else
+        docker-compose exec -T rabbitmq rabbitmqctl add_user root $RABBITMQ_PASSWORD
+        docker-compose exec -T rabbitmq rabbitmqctl add_vhost myvhost
+        docker-compose exec -T rabbitmq rabbitmqctl set_user_tags root root
+        docker-compose exec -T rabbitmq rabbitmqctl set_permissions -p myvhost root ".*" ".*" ".*"
+    fi
```
edwinmp, line 181: Not sure if this works anything like normal code... if so, can't you extract the common commands and only run the required ones within the `if` statement? i.e.
```
    if  docker-compose exec -T rabbitmq rabbitmqctl list_users | grep -q "root"; then
        log "user already exists. Skipping ..."
    else
        docker-compose exec -T rabbitmq rabbitmqctl add_user root $RABBITMQ_PASSWORD
    fi

    docker-compose exec -T rabbitmq rabbitmqctl add_vhost myvhost
    docker-compose exec -T rabbitmq rabbitmqctl set_user_tags root root
    docker-compose exec -T rabbitmq rabbitmqctl set_permissions -p myvhost root ".*" ".*" ".*"
```

config/nginx/django.conf.ctmpl
mariamiah: I added a function in the script that checks for the environment.. if the env is production it then appends it to the file...wanted to avoid manually removing it from staging
env

edwinmp: Nice 👍

akmiller01:
Tons of great work on this @mariamiah. I mainly agree with Edwin's points. If you can resolve those, and fix the conflicts with config/nginx/django.conf (might need a rebase), then we should test this on staging and get it deployed live.

Issue #809, opened 2020-06-29
Related content should not include current page


benjamincoleman:
The current page should not be able to appear in the related content section. E.g. https://devinit.org/resources/kenyas-201920-budget-and-the-big-four-agenda-a-pro-poor-analysis/

k8hughes:
This looks good to me now: http://staging.devinit.org/resources/kenyas-201920-budget-and-the-big-four-agenda-a-pro-poor-analysis/

SimonMurphyDI:
I still seem able to do this? Saved a published page here. 
http://staging.devinit.org/resources/kenyas-201920-budget-and-the-big-four-agenda-a-pro-poor-analysis/

PR #813, opened 2020-07-01
Exclude current page from related_content partial

davidebukali:
#809


edwinmp:
@davidebukali reverted your template-based solution. It reduced the number of related links when there's a duplicate. I updated the `get_related_pages` function instead.

On an unrelated note, I also fixed the `LegacyPublicationPage` template and added a `docker-compose` dev file

PR #832, opened 2020-07-09
Feature/blue green

mariamiah:


.github/workflows/workflow.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       contains('
         refs/heads/master
         refs/heads/develop
+        refs/heads/feature/blue-green
```
edwinmp, line 40: Won't this need to be removed?

Issue #826, opened 2020-07-03
Blog profile images look odd


jamesharle:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[X] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
As per Slack thread here [https://devinitorg.slack.com/archives/CCG8Y0YJG/p1593787921104000] the blog profile images are not looking as we'd want them. Ideally height of the picture would increase to accommodate larger titles, but the height would not reduce below a 1:1 aspect ratio (as it does for this blog for example: https://devinit.org/blog/what-do-trends-uks-aid-spending-nutrition-tell-us/). It would also be good if the overall blog listing page reverted to the original design.

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.

**Expected Behaviour**
A clear and concise description of what you expected to happen

**Screenshots/GIF**
If applicable, add screenshots or a GIF to help explain your problem.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Environment/Server:**
<pre><code>
- [ ] Production
- [ ] Staging
- [ ] Development
</code></pre>

**Additional Context**

Requested Delivery Date: dd/MMM/YYYY

edwinmp:
@jamesharle this is ready to test on staging http://staging.devinit.org/blog/?nav=header

jamesharle:
Thanks @edwinmp ! Is it possible to:
1) Set a 1:1 ratio as the minimum for the profile pictures as they appear on the individual blog pages? Ie. The pictures are never shorter than they are wide?
2) On the listings page, ensure that the image is aligned to the centre of the title text? Currently, for Gorik Ooms blog for example, his picture is appearing above his name and title. Ideally the vertical centre of the image would align with the vertical centre of the text box.

jamesharle:
To show what I mean:

![Capture](https://user-images.githubusercontent.com/43270042/86930468-f9185800-c12e-11ea-97fd-9a812e7c8970.JPG)

![Capture](https://user-images.githubusercontent.com/43270042/86930976-7cd24480-c12f-11ea-8142-d90d16f5469b.JPG)

edwinmp:
@jamesharle thanks for the visual aid ... I think those changes will need @benjamincoleman's input ...

benjamincoleman:
[...]

PR #835, opened 2020-07-11
Hide private blogs from the public

davidebukali:
#826


edwinmp:
Also, could you please put this in the publications PR #834 and close this one ... it also has nothing to do with Issue #826

Issue #801, opened 2020-06-23
CMS Editable Call to Action


k8hughes:
This is a new request from the comms team.

They want a pop up form, that would be editable in the CMS that allows for sign up to our mailing list.
The use case is this:  when the GHA report launches the PDF of the report will not be online, this will come a week or so later so we want users to see this pop up when they view the HTML version of the report giving them a chance to sign up to get an email when the PDF is ready.

I'm not sure how we are collecting email addresses on the website is it via a mail chimp pop up or via the CRM?

It would be good to have a quick chat about the implementation of this before someone starts to work on the issue to make sure that we get it right.

k8hughes:
@alice-mcandrew-development-initatives hopefully this sounds about right to you.  I am not sure if we can have a form from mail chimp be editiable so it might have to be a two step process ie:
Step one: CMS editable pop up appears, it has a link to the form you fill in to sign up to get the report
Step two: user see the form to fill in for mailchimp or the CRM

Note for the devs - we'd want these email addresses to go to a special list obviously...

alice-mcandrew-development-initatives:
@k8hughes Explanation looks spot on and definitely sounds sensible to talk through before starting
@AnnaHope or @jamesharle Can you advise from your perspective on how best to have this integrated with mailchimp or crm?

edwinmp:
@k8hughes @benjamincoleman  I think this may require a design

k8hughes:
@benjamincoleman it would be good to speak about a design for this, but it would take a second place after the data vis work obviously!

benjamincoleman:
@k8hughes

The question is really around how disruptive to the user viewing the report you want this to be. 

I can see two options for this: 

1. Pop up which takes over the screen, which is disruptive - the user would need to close the pop up to read the report. 
2. Call to action models designed in the page - less disruptive, but potentially less uptake. 

I would have the following content for each option: 

- Title
- Richtext
- Button

And the button I would link out to the native Mailchimp form page in a new tab, which is how the DI newsletter works currently. 

Let me know what you think.

k8hughes:
@alice-mcandrew-development-initatives @AnnaHope @jamesharle if you can confirm with Ben how you want this to function then we can get a design to work from, there are two options to choose from above.

alice-mcandrew-development-initatives:
@SimonMurphyDI ^

jamesharle:
[...]

PR #844, opened 2020-07-15
Feature/editable call to action

davidebukali:
#801


edwinmp:
This doesn't appear to be behaving and expected. All publication pages seem to have it without me having made a single change ... I suppose that's because u gave it default values. You'll have to remove that.
The behaviour (which James agreed to) is detailed [here](https://github.com/devinit/DIwebsite-redesign/issues/801#issuecomment-653538940)

The 3 pages [here](https://github.com/devinit/DIwebsite-redesign/issues/801#issuecomment-658680877) (minus the PublicationSummaryPage, whose CTA block will be managed through the PublicationPage) can have their own CTA block. For child-capable pages (PublicationPage), their corresponding children will inherit their CTA block. i.e. if a PublicationPage has a CTA block, it must appear in each child summary and chapter page.

di_website/blog/models.py
```diff
@@ -115,6 +115,12 @@ class BlogArticlePage(TypesetBodyFootnoteMixin, HeroMixin, Page):
             PageChooserPanel('internal_author_page', TeamMemberPage),
             StreamFieldPanel('other_authors')
         ], heading="Author information"),
+        MultiFieldPanel([
+            FieldPanel('call_to_action_title'),
+            FieldPanel('call_to_action_body'),
+            FieldPanel('call_to_action_button_text'),
+            FieldPanel('call_to_action_button_url'),
```
edwinmp, line 122: Give these simpler headings since they're grouped under one Panel ... you can lose the `call to action` bit ... i.e `FieldPanel('call_to_action_title', heading='Title')`

di_website/common/mixins.py
```diff
@@ -134,3 +134,13 @@ class SectionBodyMixin(models.Model):
 
     class Meta:
         abstract = True
+
+
+class CallToActionMixin(models.Model):
+    call_to_action_title = models.CharField(max_length=255, null=True, blank=True, default="Receive a pdf version", help_text="Leave title field blank to remove this call to action")
+    call_to_action_body = models.TextField(null=True, blank=True, default="We shall email you a pdf version")
+    call_to_action_button_text = models.CharField(max_length=255, null=True, blank=True, default="Signup to receive a copy in your inbox")
+    call_to_action_button_url = models.URLField(max_length=255, null=True, blank=True, default="https://us11.list-manage.com/subscribe?u=a829237ca0cf1470615c7f059&id=ce30e2af0f")
```
edwinmp, line 143: Remove the defaults, and for lines this long, put the properties on their own lines ...

di_website/publications/models.py
```diff
@@ -368,6 +368,12 @@ class Meta:
         FieldPanel('colour'),
         hero_panels(),
         StreamFieldPanel('authors'),
+        MultiFieldPanel([
+            FieldPanel('call_to_action_title'),
+            FieldPanel('call_to_action_body'),
+            FieldPanel('call_to_action_button_text'),
+            FieldPanel('call_to_action_button_url'),
+        ], heading='Call to Action Section'),
```
edwinmp, line 376: Create a `callToActionPanel` function to go with the mixin

davidebukali: Added cta panel for publications and blogs

di_website/common/mixins.py
davidebukali: Updated

di_website/blog/models.py
davidebukali: Added simpler headings

Issue #870, opened 2020-07-27
Change button from 'Publications downloads' to 'Downloads'


alice-mcandrew-development-initatives:
Please can we have the button that appears beneath the hero on publication pages changed from 'Publications downloads' to  simply 'Downloads':

![image](https://user-images.githubusercontent.com/55877063/88545220-97069080-d012-11ea-9408-48e130d9ad98.png)

(This would make more sense since we have renamed the section to 'Resources', to accommodate other formats, and also would cover us for instances where we have data but not publications for download)

edwinmp:
@alice-mcandrew-development-initatives a fix is deployed on staging ... http://staging.devinit.org/resources/p20-benin-consultation-consensus/

I also changed the label in the download list from `Publication downloads` to `Downloads`
![image](https://user-images.githubusercontent.com/5672438/88689602-563a7480-d103-11ea-80c6-69b1c7a47796.png)

alice-mcandrew-development-initatives:
@edwinmp Please can we have 'Resource downloads' at the bottom, rather than the historic 'Publication downloads' and rather than just 'Downloads' as we have now:
![image](https://user-images.githubusercontent.com/55877063/88818868-1f259b00-d1b7-11ea-95a2-3a41dab5a720.png)
(We need this to be distinct from 'Data downloads' otherwise having two headings at the bottom doesn't make sense, therefore we can't just have 'Downloads')

In an ideal world we would just have one heading at the bottom, which would be 'Downloads'. So, we would want to delete 'Data downloads' and then close the gap between the different types:
![image](https://user-images.githubusercontent.com/55877063/88818314-79722c00-d1b6-11ea-8506-5ef8fc2a1965.png)
However, I imagine this would be more work for you since it would involve rethinking how the snippets.
So, in the short-term, change to 'Resource downloads' at the bottom is fine.

edwinmp:
@alice-mcandrew-development-initatives let me look into getting the latter done ... don't imagine it's too complicate :)

edwinmp:
> In an ideal world we would just have one heading at the bottom, which would be 'Downloads'. So, we would want to delete 'Data downloads' and then close the gap between the different types:

@alice-mcandrew-development-initatives have deployed a fix for this to staging http://staging.devinit.org/resources/p20-benin-consultation-consensus/#downloads

k8hughes:
@SimonMurphyDI as Alice is off this week could you review this one for us?  Or maybe @jamesharle ?

[...]

PR #876, opened 2020-07-30
Group publication downloads under one heading

edwinmp:
Fix for [this](https://github.com/devinit/DIwebsite-redesign/issues/870#issuecomment-665728759) comment


edwinmp:
> Was this change also made in the pattern library? Or just a comms request?

Just a comms request.

Issue #793, opened 2020-06-18
Unable to upload media files


jamesharle:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[X] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
Uploading a podcast file (approx 20mb) produces the error 413 Request Entity Too Large.

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.
1. In the CMS, upload a media file
2. See error

**Expected Behaviour**
Site should accommodate files at least up to 50mb - and it should be noted that event that limit would preclude us from hosting video on the site.

davidebukali:
@jamesharle we updated permissions for the media folder, kindly test this on staging http://staging.devinit.org when you get some free time

jamesharle:
Thanks @davidebukali! Can see that the specific issue flagged here is resolved as I've confirmed I'm now able to upload files. My testing so far tells me that the max size for these is around 50mb, which will be suitable for our audio requirements.

jamesharle:
Hi! Have just tried this with an 11MB file today, and getting the same error as previously, that the entity is too large :(

sarahharries66:
@edwinmp Just tagging to let you know that James has reopened this ticket

edwinmp:
Sure ... won't be in tomorrow, but will have a look at it on Monday.

edwinmp:
There's a server limit configured to 10MB ... changed that to 100MB, but will need to deploy the change and restart before it reflects ... will get time late in the night for that.

edwinmp:
@akmiller01 to fix this issue, I need to have [this](https://github.com/devinit/DIwebsite-redesign/pull/883/commits/cf8b5903bc658f1c077095aeed6848c69706fa55) change on the live server, but according to @mariamiah every attempt to manually change it and restart has been somehow reverted after the restart ... any ideas?

edwinmp:
Hi @jamesharle @sarahharries66 ... this issue is now fixed ... please test to verify.

PR #883, opened 2020-08-17
fix failing ci

mariamiah:


.github/workflows/workflow.yml
```diff
@@ -35,6 +35,7 @@ jobs:
       contains('
         refs/heads/master
         refs/heads/develop
+        refs/heads/fix/failing-ci
```
edwinmp, line 38: Is this staying? I tend to spot these in each of your CI PRs ...

config/nginx/django_https.ctmpl
```diff
@@ -10,7 +10,7 @@ server {
 
     server_name www.devinit.org devinit.org;
 
-    client_max_body_size 10M;
+    client_max_body_size 100M;
```
edwinmp, line 13: I added this ... it's supposed to fix [this](https://github.com/devinit/DIwebsite-redesign/issues/793#issuecomment-671249922) issue

edwinmp (approved):

Issue #887, opened 2020-09-01
Block quote functionality missing from Short publication and Publication page templates in Resources


CeliaWarin:
## I'm submitting a ...

<pre><code>
[X] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
In resources, the Short Publication template and Publication Page (including child pages Publication Summary, Publication Appendix and Publication Chapter and Publication Foreword) are missing the Block quote:

![image](https://user-images.githubusercontent.com/70264269/91853907-cdab7880-ec5a-11ea-920f-911051ec5bbf.png)

This is available on other pages such as blogs:

![image](https://user-images.githubusercontent.com/70264269/91854043-051a2500-ec5b-11ea-9f2a-727d02d79517.png)

According to the pattern library, this should also be available on these pages: 
http://development-initiatives.surge.sh/page-templates/14-1a-publication-short-form
http://development-initiatives.surge.sh/page-templates/14-1-publication-long-form
http://development-initiatives.surge.sh/page-templates/14-2-publication-appendix

This means we're currently having to format quotes using bold and italics in the Wysiwyg editor which leads to a lack of consistency across the website. 

For example, a quote on the pubs pages styled in the Wysiwyg editor:
![image](https://user-images.githubusercontent.com/70264269/91854097-20853000-ec5b-11ea-9215-8f2a8d647698.png)

A correctly styled quote on another page using the Block quote:
![image](https://user-images.githubusercontent.com/70264269/91854222-50343800-ec5b-11ea-84ea-c70682e11582.png)

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.
1. Go to 'https://devinit.org/admin/pages/352/ - Pages > Homepage > Resources'
2. Click on 'Add child page'
3. Click on '+Short Publication' 
4. Scroll down to 'Content'
5. You will see on the content menu that there is no option for Block quote

**Expected Behaviour**
When adding content blocks to a Short publication page or Publication page (including child pages Publication Summary, Publication Appendix and Publication Chapter and Publication Foreword), the option to create a Block quote should be available, i.e. 

![image](https://user-images.githubusercontent.com/70264269/91854043-051a2500-ec5b-11ea-9f2a-727d02d79517.png)

mariamiah:
[...]

PR #889, opened 2020-09-02
add blockquote functionality to short publication and publication pages

mariamiah:
#887 
### changes
- adds blockquote functionality to the short publication and publication pages
blockquote


edwinmp:
Did you test on the actual pages this will affect and made sure it appears as expected? One vital feature is that the `blockquote` must take on the colour of the parent publication. This is likely working by default, but need to know that it's been tested.

edwinmp:
Also, I have a strong feeling that this feature should come with migrations.

mariamiah:
> Did you test on the actual pages this will affect and made sure it appears as expected (not just on the admin side)? One vital feature is that the `blockquote` must take on the colour of the parent publication. This is likely working by default, but need to know that it's been tested.

Yes, tested with the appendix, chapter, foreword, summary and short publication pages and they all appear as expected

edwinmp:
image

This doesn't match the design ... these blockquotes are too wide ... yes, that's important ... comms reviewers are that particular, so please make sure it matches the designs in the pattern library.
I suggest u don't edit this block's template as it's already being used as it is elsewhere ... provide another template that fits this use case. https://docs.wagtail.io/en/v2.9/topics/streamfield.html#template-rendering

di_website/publications/blocks.py
```diff
@@ -49,3 +48,16 @@ class AudioMediaStreamBlock(StreamBlock):
...
+    Custom `StructBlock` that allows render text in a blockquote element
+    """
+    text = TextBlock()
+    source = TextBlock(required=False, help_text='Who is this quote acredited to?')
+    center = BooleanBlock(default=False, required=False)
+
+    class Meta:
+        icon = 'fa-quote-right'
+        template = 'blocks/publication_blockquote.html'
+
```
edwinmp, line 63: This is an entirely new block, which is redundant ... u only need to specify a different template when using the already existing block i.e `BlockQuote(template='path/to/template')`

edwinmp (changes requested):

edwinmp (approved):
Fixed the uncentered view and added the missing migrations

Issue #886, opened 2020-08-28
display of 3 digit notes


CeliaWarin:
## I'm submitting a ...

[X ] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
We need to add additional space between note and text for 3 digit notes so consistent with 1 and 2 digit notes and legible  

![image](https://user-images.githubusercontent.com/70264269/91550306-44c1d380-e920-11ea-8825-46dfc05cc2a5.png)

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.
1. Go to 'https://devinit.org/resources/socioeconomic-impact-of-covid-19-in-uganda/'
2. Scroll down to Notes

**Expected Behaviour**
Text is aligned and there is a space between note and text

**Desktop (please complete the following information):**
n/a 

**Smartphone (please complete the following information):**
n/a

edwinmp:
@CeliaWarin please test a fix for this [here](https://dev.devinit.org/resources/socioeconomic-impact-of-covid-19-in-uganda/)

CeliaWarin:
Hi Edwin,

I’ve tested and works fine thanks.

Best
Celia

CeliaWarin:
Hi Edwin,

I’ve tested and published the page and seems to be working fine – thanks.

Best
Celia

PR #898, opened 2020-09-08
Add padding for 3 digit footnotes

davidebukali:
This affects footnotes with 3 digits, that is > 99. Testing with dokku deployment seems ideal to avoid having to create 100 footnotes on your local environment. I copied `wysiwyg` content from the [devinit.org](https://devinit.org/admin/pages/2138/edit/) website to test on `localhost`.
[Issue details](https://github.com/devinit/DIwebsite-redesign/issues/886#issue-687971345)

### DIFF
--- a/di_website/static/css/di_website.css
+++ b/di_website/static/css/di_website.css
@@ -83,6 +83,6 @@
   stroke: transparent !important;
 }
 
-.three-digit-footnote {
+.footnotes__item {
   padding-left: 10px;
 }
--- a/di_website/static/css/di_website.css
+++ b/di_website/static/css/di_website.css
@@ -82,3 +82,7 @@
 .hovertext path {
   stroke: transparent !important;
 }
+
+.three-digit-footnote {
+  padding-left: 10px;
+}
--- a/di_website/templates/publications/footnotes.html
+++ b/di_website/templates/publications/footnotes.html
@@ -7,7 +7,11 @@
                 <h2 class="heading type-l type-l--trailer">Notes</h2>
                 {% for item in page.footnotes_list %}
                 <ul class="footnotes__list">
+                {% if forloop.counter <= 99 %}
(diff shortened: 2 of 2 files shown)


di_website/templates/publications/footnotes.html
```diff
@@ -7,7 +7,11 @@
                 <h2 class="heading type-l type-l--trailer">Notes</h2>
                 {% for item in page.footnotes_list %}
                 <ul class="footnotes__list">
+                {% if forloop.counter <= 99 %}
                     <li class="footnotes__item">
+                {% else %}
+                    <li class="footnotes__item three-digit-footnote">
+                {% endif %}
```
edwinmp, line 14: Inline this `<li class="footnotes__item {% if forloop.counter > 99 %}three-digit-footnote{% endif %}">`

edwinmp (changes requested):
Looks like you're also into dokku ... sadly, no dokku here ...

di_website/static/css/di_website.css
```diff
@@ -82,3 +82,7 @@
 .hovertext path {
   stroke: transparent !important;
 }
+
+.three-digit-footnote {
```
edwinmp, line 86: Actually, add this padding to the `footnotes_item` class, that way they're all aligned
Instead of this:
image

We get this:
image

It keeps them aligned all the way to 999, plus u don't need the template condition.

edwinmp (changes requested):

edwinmp:
> Testing with dokku deployment seems ideal to avoid having to create 100 footnotes on your local environment.

All you had to do was `inspect element` and change whatever footnote u had to any value, right in the HTML ... no need to deploy or hunt for DBs.

davidebukali:
Updated html to remove conditional and use existing `css` class

edwinmp (approved):

PR #914, opened 2020-10-02
Scrub FigurePage from website, amend topic tags

akmiller01:


edwinmp (approved):

akmiller01:
Tested this out locally. But can we also trial on dev before deploy @edwinmp ? Figure pages were woven into quite a few pieces.

Issue #901, opened 2020-09-15
Extra features on event previews


sarahharries66:
**Is your feature request related to a problem? Please describe.**
Currently our event previews (the previews of event pages that appear in listings or as related content on other pages) are pretty simple and just have title, date and time - and the time can only be set in whole hours. This isn't ideal because we often have events that don't start on the hour, or are in different timezones. 

**Describe the solution you'd like**
We'd like a free-text box of around 160 characters that would pull through above the date and time to add more detail (eg timezones) or specificity (eg half or quarter hour increments).

**Describe alternatives you've considered**

**Additional context**
![image](https://user-images.githubusercontent.com/69594823/93195941-b469f800-f741-11ea-94ee-2f9e28673ce0.png)

**Requested Delivery Date:** dd/MMM/YYYY

edwinmp:
@davidebukali your implementation here doesn't meets the requirements in the issue. Timezone was an example of just one problem that this would solve. The requested change was to replace the date/time fields with a free-text field.

davidebukali:
@sarahharries66 would you prefer one text field for both date and time information or two separate fields ? Also kindly give an example of text to be written for date and time as well as any additional detail needed.

jamesharle:
@davidebukali @sarahharries66 Overall, we need to be able to display (in addition to the name and date of the event and a thumbnail of the hero image): the exact time of the event (including the timezone), and the location of the event. So this could be handled with one freetext field of 160 characters, and would contain something like:

12.30-13.35 EAT
Kenya

We'd want that free text to appear below the date, which currently is mandatory on the CMS, and to be able to add line breaks to it manually. I attach a mockup below.

Also, just to be clear, this would be our minimum requirements - ideally we'd like the time function to include all increments in the CMS (so that we can have start time at 12.13, for example, or 14.56), a free text field to add a timecode (BST, for example, or EAT), and the free text field for location. But this solution of one larger textbox has been suggested to put less pressure on capacity for now!

![image](https://user-images.githubusercontent.com/43270042/97329420-ec9b4500-186e-11eb-8ac3-1a2a8717a6bf.png)

davidebukali:
@jamesharle @sarahharries66 this is now deployed to test on staging [events](https://dev.devinit.org/admin/pages/623/)

alice-mcandrew-development-initatives:
@davidebukali @edwinmp this meets the minimum requirement, so thank you very much!
[...]

PR #921, opened 2020-10-14
Feature/event previews

davidebukali:
From #901 ..."to add more detail (eg timezones) or specificity (eg half or quarter hour increments)."


akmiller01 (approved):
Looks good. I'm tempted to use the built-in timezone in the date format, but knowing how this will be used, the CMS users will want to specify all sort of different timezones.

Issue #913, opened 2020-10-01
Change to how pages show as related content 


SimonMurphyDI:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[ ] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
A clear and concise description of what the issue is.

Hiya! When we select a page as related content, it shows the date that page was published. So this looks different from resources (and so it doesn't look out of date) please can we change it so pages don't pull dates through?
![image](https://user-images.githubusercontent.com/22001094/94781049-36eceb80-03c1-11eb-8fc4-3fc6f942dee7.png)
Thanks
Simon

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.

**Expected Behaviour**
A clear and concise description of what you expected to happen

**Screenshots/GIF**
If applicable, add screenshots or a GIF to help explain your problem.

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Environment/Server:**
<pre><code>
- [ ] Production
- [ ] Staging
- [ ] Development
</code></pre>

**Additional Context**

Requested Delivery Date: dd/MMM/YYYY

davidebukali:
@SimonMurphyDI is this the preferred behavior only on [publication pages](https://devinit.org/resources/uganda-disability-data-landscape-economic-inclusion-persons-with-disabilities/)  or across all resources (AudioVisualPages, LegacyPublications etc)

SimonMurphyDI:
@davidebukali - all pages please! :)

davidebukali:
@SimonMurphyDI alright, thanks :)

davidebukali:
@SimonMurphyDI when you get time, this is ready to test on [staging](https://dev.devinit.org/)

SimonMurphyDI:
Hi @davidebukali - the date has been removed for all pages, resources, blogs etc. It's important we keep it for resources and blogs, etc ... the only place we don't want a date to pull through is on pages, as this is content that should be up to date (and so it doesn't matter when it was published). It does matter to the user when a report was published.

edwinmp:
@SimonMurphyDI that's confusing as they're all pages ... could you please share links of pages u want it to show and those u don't want it showing.

edwinmp:
My understanding is that you only want it to show for resources, blogs, news, events (am I missing any ... no e.t.c) ...

[...]

PR #926, opened 2020-10-17
Feature/update related content

davidebukali:
Hide date text for related content pages attached to all resources #913


di_website/templates/includes/cards/card-blog.html
```diff
@@ -10,7 +10,9 @@
 
     <div class="card__body">
       {% if blog.published_date %}
-        <span class="card__meta">{{ blog.published_date|date:"j F Y" }}</span>
+        {% if content_type != 'resources' %}
```
edwinmp, line 13: Use a more specific variable that doesn't result in confusing usages ... i.e `hide_date` ... then check `if not hide_date`

Besides the actual card templates, the only other change should be in `related_content.html` where you'll set `hide_date` to `True`

edwinmp (changes requested):

di_website/templates/includes/cards/card-blog.html
davidebukali: @edwinmp added `hide_date` to publication models and templates

di_website/publications/models.py
```diff
@@ -471,6 +478,12 @@ class Meta:
 
     colour = models.CharField(max_length=256, choices=COLOUR_CHOICES, default=RED)
 
+    hide_date = models.BooleanField(
+        default=True,
+        verbose_name="Hide Related Links Date",
+        help_text="Should we hide the date on related links items ?",
+    )
```
edwinmp, line 485: Dude? Why? ... I added my comment on a template file but u added a model field of it ... overkill ...

edwinmp: I don't get it ... why would u implement a clearly terrible idea?? Since u misunderstood, you should have at least questioned my sanity cos this is a bad idea ... all I wanted was that template variable `content_type` to be renamed `hide_date` ... then I actually mentioned that "the only other change" would be in `related-content.html` ... I think you were the drunk one here.

edwinmp (changes requested):
This has spiralled ... I'll fix it.

edwinmp (approved):

di_website/publications/models.py
davidebukali: My bad, will get a second opinion next time. Thought you were refering to `configuration over code` so that users could change `hide_date` in the admin for each of the resources pages.

PR #935, opened 2020-11-03
Feature/event previews

davidebukali:
Add a text field to hold event time, timezone and location #901


di_website/events/migrations/0035_auto_20201103_1256.py
```diff
@@ -0,0 +1,27 @@
+# Generated by Django 3.1 on 2020-11-03 12:56
+
+from django.db import migrations
+
+
+def combine_time_and_location(apps, schema_editor):
+    EventPage = apps.get_model('events', 'EventPage')
+    for event in EventPage.objects.all():
+        event.time_and_location = '%s - %s %s %s' % (event.start_time.strftime("%H:%M"), event.end_time.strftime("%H:%M"), '\n', event.location)
```
edwinmp, line 9: Getting this error when running migrations
image

Probably cos you're assuming `start_time` & `end_time` are always provided.

di_website/events/models.py
```diff
@@ -28,10 +28,7 @@ class EventPage(TypesetBodyMixin, HeroMixin, Page):
 
     start_date = models.DateField(default=datetime.now)
     end_date = models.DateField(default=datetime.now)
-    start_time = models.TimeField(blank=True, null=True)
-    end_time = models.TimeField(blank=True, null=True)
-    timezone = models.CharField(max_length=160, help_text='Timezone information for event', blank=True, null=True)
-    location = models.CharField(max_length=100, help_text='Physical location of event', blank=True, null=True)
+    time_and_location = models.TextField(max_length=160, help_text='Time and location information for event', blank=True, null=True)
```
edwinmp, line 31: image

I don't think it's that much work to create two free text fields, one for time & one for location as requested in the text above ... that way we can at least preserve the icons. The location one was already there, so it's just a matter of a free-text time field. Would also cut down on your data migration hustles.

edwinmp (changes requested):

di_website/events/models.py
davidebukali: Updated with two fields for time and location

di_website/events/migrations/0035_auto_20201103_1256.py
davidebukali: Now checking that `start_time` and `end_time` are set before calling `strftime()`

edwinmp (approved):

Issue #937, opened 2020-11-20
Extra information on our podcast pages 


sarahharries66:
We're looking to add some more features to our podcast pages to help give out a bit more information on the content and participants etc. of the podcast. 

we're looking to add the following: 

1. A label in the header specifying publication date
2. A label in the header specifying the resource type (podcast)
3. Any information pulling through on participants
4. The ability to add a country of relevance (this is missing for Webinar and Presentation types also)

For a good idea, see the old page set up: 
![fffixture - Google Chrome 20_11_2020 14_46_34_LI](https://user-images.githubusercontent.com/69594823/99813262-5db4dc00-2b3f-11eb-953b-e5aa07068ee9.jpg)

no set request date on these but it would be great to get an idea when you'll have capacity to make these changes.
**Requested Delivery Date:** dd/MMM/YYYY

Tagging: @jamesharle

davidebukali:
@benjamincoleman looks like details from 1 to 4 can be added by the dev team; am not sure about the country, might this need  design input ?

benjamincoleman:
Morning @davidebukali 

Here's my input on each task: 

1. A label in the header specifying publication date
2. A label in the header specifying the resource type (podcast)

Both 1. and 2. are correct, you should have this content for all publication types: 

![Screenshot 2020-12-03 at 10 28 08](https://user-images.githubusercontent.com/64877/100997562-4e2b8f00-3552-11eb-8c1d-5f7fd9aa1268.png)

3. Any information pulling through on participants

I would recommend adding an "about the authors / participants" link to the authors module. This would then scroll the user down, to a new module at the end of the content which can talk either about the authors, or about the participants. I can produce a sketch if you would like to show what I mean. 

4. The ability to add a country of relevance (this is missing for Webinar and Presentation types also)

This is a development task, but yes, there should be a country field for it - otherwise they will not be discovered when a user fitlers resources by a specific country.

davidebukali:
Hi @benjamincoleman thank you for the feedback; Since we both agree on 1 and 2, I have two comments for 3 and 4.

In task 3, there's a card U.I for publications with a list of names for authors; from your description, podcast pages appears to work differently from publications. It would be nice to see a sketch to guide development.

![authors](https://user-images.githubusercontent.com/1607276/101064419-07b35000-35a5-11eb-902f-066abf88c632.png)

[...]

PR #942, opened 2020-12-04
Feature/add podcast details

davidebukali:
Add details to `Audio Visual Pages` #937

### DIFF
--- a/di_website/templates/includes/partials/multi-authors.html
+++ b/di_website/templates/includes/partials/multi-authors.html
@@ -3,7 +3,7 @@
 <div class="m-authors">
     <h3 class="m-authors__title">
       {% if title and authors|length > 1 %}
-        Participants
+        {{ title }}
       {% elif title %}
         Participant
       {% elif first_author and authors|length > 0 %}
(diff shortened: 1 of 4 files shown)


di_website/templates/includes/partials/multi-authors.html
```diff
@@ -2,7 +2,11 @@
 
 <div class="m-authors">
     <h3 class="m-authors__title">
-      {% if first_author and authors|length > 0 %}
+      {% if title and authors|length > 1 %}
+        Participants
```
akmiller01, line 6: Works well and looks good. Would it make sense to make this line `{{ title }}` instead of `Participants` so it can take the title argument from `audio_visual_media.html`?

akmiller01 (changes requested):
Just one question on line 6/8 of multi-authors.html

di_website/templates/includes/partials/multi-authors.html
davidebukali: @akmiller01 great idea, its now updated

akmiller01 (approved):

PR #943, opened 2020-12-05
Update environment variables for git workflow

davidebukali:
Update environment variables in git workflow.

More info [here](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/)

And [here](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files)


akmiller01 (approved):
:+1:

PR #992, opened 2021-04-21
Update | Advanced Charts

edwinmp:
- Alternative text
- Extra rich text features for instruction text


akmiller01 (approved):
Looks good!

akmiller01:
Maybe worth removing `MINIMAL_RICHTEXT_FEATURES` from constants.py, as I don't think it's used anywhere else.

PR #995, opened 2021-04-26
Add data systems charts for the board

davidebukali:
Create charts described in [issue 7](https://github.com/devinit/org-dashboard-data/issues/7)


src/dashboard/utils/dashboards/dataSystems.ts
```diff
@@ -1,3 +1,114 @@
-import { DashboardGrid } from '../../../utils/types';
+import { getAggregatedDatasetSource } from '..';
+import { DashboardData, DashboardGrid } from '../../../utils/types';
+import { getEventHandlers, grid } from '../chart';
+
+const colours = ['#FF9A5C', '#FFA770', '#FFB485'];
```
edwinmp, line 5: Pick a colour palette from [this](http://archive.devinit.org/wp-content/uploads/2018/03/colours.html), one that's not been used in any of the other sections ...

edwinmp (changes requested):

src/dashboard/utils/dashboards/dataSystems.ts
edwinmp: Has to be monochromatic, but with good contrast between the colours ... pick at least 5

davidebukali: Picked purple from the pallete

edwinmp (approved):

Issue #999, opened 2021-05-10
no cookies pop up when you visit the site?


SimonMurphyDI:
## I'm submitting a ...
<pre><code>
[x] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
When you visit our website you aren't informed what cookie settings the website has, and what options you have to change these.  

**To Reproduce**
1. Go to the website 

**Expected Behaviour**
On our previous website, when you arrived on the domain, you were given a pop-up that explained our cookie policy. I think we're legally required to explain what cookies the site is using. It would be good to let visitors know if they have the ability to amend the cookies we use. This may require dev if we don't have a pop up with the ability to amend cookie settings. I imagine we'd need this to work on mobile and desktop 

Richard Y has said he's writing text to sit in this box 

We do have a 'notice heading' on the homepage here - could we use this?
http://development-initiatives.surge.sh/page-templates/01-general-page

**Requested Delivery Date**:

This feels very urgent to me - Richard hasn't provided an estimated delivery date for his text

edwinmp:
Hi @SimonMurphyDI, besides the google tag manager, do we have any other trackers that use cookies?

A simple message explaining that we use cookies, maybe with a link to a Wagtail page with our cookie policy would do. If you need one that gives users control over which cookies they allow, then I'll need to know what the cookies are first and see how to get it done.

SimonMurphyDI:
Thanks @edwinmp! I know we use google tag manager and mouseflow. I don't know of any more than that, but think it would be worth confirming this in the chat - will do that

edwinmp:
Hi @SimonMurphyDI, do we have the text that'll sit in this box? I'll be best to have the basic implementation done & deployed while we look into giving the user more control.

LouiseSummerlring:
Hello @edwinmp! I'm helping Simon out with some website stuff at the moment. We do indeed have the text for the cookies pop-up box, and the new and improved cookie policy itself. The policy document is attached, and the text is below.

We use cookies on our website to help us understand how you use the site and to make improvements. These may be used to track your visit to our website or to enable us to personalise your browsing experience, and they may be shared with Google Analytics so that your data can be processed. For more information, please visit our Privacy and Cookie Policy here.”

On the word 'here', we should link to the policy document.
[...]

PR #1014, opened 2021-07-02
Feature | cookie popup

davidebukali:
Add notice for users to read privacy and cookie policy #999 

Displays content created under the first `cookie notices` snippet


edwinmp (changes requested):
Great work getting this done so fast.

The notice thingie always shows, even when there's no cookie notice added yet. It should depend on at least the text or body being present.
image

Both heading and body should not be required as one may want to use or the other, or both
image

Interesting way of handling the link ... lets hope they won't ever want it to appear in the middle of the text lol ...

edwinmp (approved):

PR #1043, opened 2021-09-21
Make all sh scripts executable

davidebukali:
Fix for [faulty](https://github.com/devinit/DIwebsite-redesign/runs/3663426471) deployment


edwinmp:
It was a permissions issue as indicated in the error ... since it wasn't there before, it probably originated from some action on the server. Anyway, I've fixed it on the server and ran the broken deployment. Closing this.

Issue #1060, opened 2021-09-24
New Table A2 on Covid Tracker Page


bill-anderson:
Add a new interactive table below Table A1 on the Covid Tracker Page

- Title TBC
- The data for this table will be updated from DDW by analysts on a monthly basis to a CSV in GitHUB at the same time as the data for the main visualisation is updated.
- This CSV contains 6 columns: Reporting Organsation Reference, Name, Organisation Type, Transaction Type, Value $m, YYYYMM 
- The row labels in the table are populated from the Name column in the csv
- The column labels are populated from the YYYYMM column in the csv
- The value cells are populated from the Value $m column in the csv
- Two filter dropdowns limit the display: Organisation Type (values are Bilateral, IFI and Multilateral) and Transaction Type (values are Commitment and Spend)
- See Pivot table in attached file

[Tracker metadata.xlsx](https://github.com/devinit/ddw-analyst-ui/files/7212207/Tracker.metadata.xlsx)

k8hughes:
We will implement this as something simpler than a pivot table, instead just a table with drop downs powered by a CSV uploaded to github.

@akmiller01 has ideas about this

k8hughes:
@edwinmp to speak to Bill and Dean about this and include into the next sprint.

k8hughes:
@dean-breed will need to have completed the quality check work first for this.  SO revision to the post above, we can't work on this just yet, @dean-breed to let us know when unblocked

akmiller01:
Commenting to unstale. This'll require a bit more code than A1, but not much.

edwinmp:
[Here's](https://dev.devinit.org/admin/pages/2456/edit/) the pivot table, configured through a StreamField (for the most basic features). I'll add a toggle for the size on this one as well.
Let me know if there's anything else you need changed or added.

![image](https://user-images.githubusercontent.com/5672438/150333235-c2abd495-e961-4936-a6a8-36968e77c3f2.png)


![image](https://user-images.githubusercontent.com/5672438/150333467-4ac5ab7f-6a65-4459-9d60-c0429c8cdeb4.png)

dean-breed:
@bill-anderson do you think this should be All flows or just ODA?

bill-anderson:
Definitely all flows.

dean-breed:
@bill-anderson got it!

dean-breed:
@bill-anderson One final thing - do we want it by agency (like the example above) or by grouping (like the chart)?

[...]

PR #1088, opened 2022-01-10
Feature | Pivot Table

edwinmp:
Ticket #1060


ThatcherK:
@edwinmp ready fro review

edwinmp:
> @edwinmp ready fro review

@ThatcherK I've updated the CI to test the TS/JS ... that way if your local setup is not proper, u won't miss anything... there's plenty of errors in the current code. Look into those while I do a functional test

edwinmp (commented):
- Make the labels bold i.e `Row Labels`, `Grand Total`
- Give the `Grand Total` column label a `white-space: nowrap` style so that all the text is on a single line
- Format the numbers i.e instead of 1245, it appears as 1,245
- For the cell values, round up with no decimal places, and this is what you should use in the total i.e for 0.623, the display value should be 1

ThatcherK:
@edwinmp  ready for another review

PR #1101, opened 2022-01-28
Multiple condition row  highlighting on pivot-table

ThatcherK:
Ticket #1060


di_website/visualisation/models.py
```diff
@@ -189,6 +191,62 @@ class RawCodePage(CodePageMixin, Page):
...
+    GREEN = '#2D5A27'
+    COLOUR_CHOICES = (
+    (RED, 'Red'),
+    (BLUE, 'Blue'),
+    (PINK, 'Pink'),
+    (YELLOW, 'Yellow'),
+    (ORANGE, 'Orange'),
+    (PURPLE, 'Purple'),
+    (GREEN, 'Green')
+    )
```
edwinmp, line 218: Don't limit the colour ... leave it as a blank field... give it the default as I requested in the instructions. I know these peeps ... they're too fickle and I'd rather not have to code anymore just to accommodate them. If you're set on these colours, add them in the `help_text` as the suggested options

src/components/PivotTable/PivotTable.tsx
```diff
@@ -28,8 +30,8 @@ export const StyledPivotTableHeader = styled.th<{ column: string }>`
 export const FilterWrapper = styled.div`
   padding: 1rem 2rem 2rem 2rem;
 `;
-export const TableRow = styled.tr<{ highlight: boolean }>`
-  background: ${(p) => (p.highlight ? '#ffb3b3' : 'none')};
+export const TableRow = styled.tr<{ highlight: string }>`
```
edwinmp, line 33: Give it a different name, one that says what it is ... a colour

src/components/PivotTable/utils/index.ts
```diff
@@ -158,21 +160,28 @@ export const addCommas = (rows: string[][]): string[][] => {
...
     case 'eq':
-      return cellValue === compareValue;
+      return value === compareValue;
     default:
       return false;
   }
 };
+
+export const rowHighlightChecker = (highlightedRows: HighlightedRow[], row: string[]): string => {
+  const myChecker = highlightedRows.filter((item) => item.label === row[0]);
```
edwinmp, line 183: Use `find` ... it returns a single value or null

edwinmp (changes requested):

di_website/visualisation/models.py
ThatcherK: I was thinking that if I left it blank, there would be room for error, felt safer knowing the color before hand

edwinmp: It's manageable error, won't break anything & the flexibility it provides is worth it ... I'd rather not have comms or Bill come to me every few weeks asking me to add a new colour.

ThatcherK: Okay

edwinmp: If you really want to be on the safe side, you can provide a link to a hex colour resource like this one https://hexcolorcodes.org/ in the help_text

edwinmp: Otherwise I trust the people using this to know their way around HTML colours.

ThatcherK:
@edwinmp changes made as requested

Issue #1084, opened 2022-01-07
Making some details on role profiles optional to avoid gaps on the vacancy listing page


SimonMurphyDI:
**Is your feature request related to a problem? Please describe.**
We're currently getting a lot of role profiles for the website that don't have 
- closing dates
- salary scales
This means we have blank categories on[ the vacancies listing page](https://devinit.org/work-for-us/vacancies/), which doesn't look good. 

**Describe the solution you'd like**
It would be great if those two categories were optional, or did not appear on the vacancies listing page if there was no information entered. 

**Describe alternatives you've considered**
I don't think there are any good alternatives.

**Additional context**

Here's how a current vacancy looks - almost appears there's an error.

![image](https://user-images.githubusercontent.com/22001094/148544052-1b9a222f-de0b-4997-b158-f14e0883a28a.png)

**Requested Delivery Date:** 
Happy for you to do this when you can, as what's there is not wrong, it just doesn't look good.

Thank you!

edwinmp:
@SimonMurphyDI this is fixed on dev - please review

SimonMurphyDI:
thanks @edwinmp

PR #1105, opened 2022-02-07
Feature | Optional Vacancy Role Profiles

pauline-n:
Making some details on role profiles optional to avoid gaps on the vacancy listing page #1084

### DIFF
--- a/di_website/templates/includes/cards/card-job.html
+++ b/di_website/templates/includes/cards/card-job.html
@@ -6,8 +6,8 @@
         <h3 class="card__title"><span>{{ vacancy.title }}</span></h3>
         <span class="card__meta">{{ vacancy.location }}</span>
         <ul class="card__list">
-            <li><span>Salary scale: </span>{{ vacancy.salary_scale }}</li>
-            <li><span>Closing date: </span>{{ vacancy.application_close|date:"j F Y" }}</li>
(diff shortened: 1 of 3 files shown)


di_website/templates/includes/cards/card-job.html
```diff
@@ -6,8 +6,8 @@
         <h3 class="card__title"><span>{{ vacancy.title }}</span></h3>
         <span class="card__meta">{{ vacancy.location }}</span>
         <ul class="card__list">
-            <li><span>Salary scale: </span>{{ vacancy.salary_scale }}</li>
-            <li><span>Closing date: </span>{{ vacancy.application_close|date:"j F Y" }}</li>
+            <li><span></span>{{ vacancy.salary_scale }}</li>
+            <li><span></span>{{ vacancy.application_close|date:"j F Y" }}</li>
```
edwinmp, line 10: Nope ... why have you removed the labels? hmm... are you aware that you can use conditional operators in Django templates??

edwinmp (changes requested):

di_website/templates/includes/cards/card-job.html
pauline-n: Ohh yes I see it now, removing the labels doesn't really solve the issue. Yes to the conditional operators, let me resolve this and push the changes, thank you.

edwinmp (approved):

Issue #1117, opened 2022-03-03
Handle ValueError resulting from null character injection in resources and dataset search/filter strings


edwinmp:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[x] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
On resources and dataset pages, malicious search or filter requests that included percent-encoded Unicode NULLs (%00) raised a ValueError exception: `A string literal cannot contain NUL (0x00) characters`. I'm not sure what the best way to fix this is, but there were two options so far...

1. Catch and handle these values errors to avoid a 500 error
2. Create a custom middleware that implements the solution [here](https://github.com/wagtail/wagtail/pull/4652/files)

I'm leaning towards option 2 ... you can find a custom middleware implementation for the IATI website [here](https://github.com/IATI/IATI-Standard-Website/pull/610/files)

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.
1. Go to www.devinit.org/resources/?country=bangladesh&q=the&topic=1%00%C0%A7%C0%A2%252527%252522&types=factsheet
5. Add the value of topic above into any of the other fields on the resources or dataset pages for the same result

**Expected Behaviour**
No 500 error for this kind of attack

Requested Delivery Date: dd/MMM/YYYY
ASAP as the most recent attack resulted in over 100 error emails. Very annoying.

PR #1118, opened 2022-03-10
Fix/value error null character search

ThatcherK:
ticket #1117


edwinmp (changes requested):
This solve it directly for those pages ... as for every subsequent page we want to add a filter to, we'd have to do this ... so nope. You'll need to create a custom mixin as indicated in the ticket, cos that solves the problem permanently & we won't need to do any more coding.

edwinmp (approved):
Just the single item.

Issue #1127, opened 2022-03-29
Switching from Google's Universal Analytics to GA4 


SimonMurphyDI:
@akmiller01 

As per our conversation, a space to consider the implications of this shift for the DI and GNR properties. 
Known needs:
- update our account
- save our previous data
- tag manager (TBC)

edwinmp:
@SimonMurphyDI how soon do you need this?

akmiller01:
@edwinmp I was thinking that as if we can get GA4 in place by July 2022, then even if the data from UA to GA4 is incomparable, we'll have 1 year of GA4 data for future comparisons by the time UA gets shut off.

SimonMurphyDI:
@akmiller01 we might want to plan the transfer around reporting for key outputs like the GHA report (to avoid a chunk of missing analytics at the wrong moment) - or could we run both old and new at the same time?

akmiller01:
We should be able to run both old and new simultaneously.

Next steps for this is probably go to ahead and make a GA4 for devinit so you can assess whether it's something you'd like to use in the future, or whether you'd like to consider non-Google trackers.

What I'm trialing for IATI is this self-hosted analytics service called Umami. Good information and a demo available here: https://umami.is/

pauline-n:
@akmiller01 this has been done. Kindly. test to see if all is as required. I have also exported and backed up the UA data to google sheets - this can found with the email's google sheet files with the title UA Data Backup DI Website.

Thank you.

akmiller01:
Hi @SimonMurphyDI , Pauline has kindly setup a parallel GA4 property for the DI website. You'll find it on the analytics account under "DI Website - GA4." When your team has a moment, could you take a look and see if it fulfills your analytical needs?

Personally I've found that they changed GA4 pretty drastically from UA, such that it's much more focused on business needs like conversions and less on events/page views.

edwinmp:
Closing this ... @SimonMurphyDI reopen or create a new ticket if there's anything left to do.

PR #1143, opened 2022-05-10
Create a GA4 account for DI website

pauline-n:
Issue #1127


edwinmp:
If this PR is not ready for review (as it appears), please don't request for a review ... also, create it as a draft PR.

pauline-n:
@edwinmp it actually is ready, I actually didn't need to change anything. the same google tag manager in the <head> is connected to the GA4 account.

PR #1144, opened 2022-05-10
Update Readme with the working .env url

pauline-n:
The url env was not working, updating it with he current url that is working


.gitignore
```diff
@@ -129,3 +129,5 @@ node_modules/
 
 /src/assets/**/js/*.js
 /src/assets/**/js/*.LICENSE.txt
+
+package-lock.json
```
edwinmp, line 133: Nope ... please don't ignore package-lock.

edwinmp (changes requested):

.gitignore
pauline-n: Done

edwinmp (approved):

Issue #1151, opened 2022-06-01
Add Twitter conversion tracking


Tim-Molyneux:
Hi all - I think it would be good idea to add conversion tracking for twitter to our website, which would help us better track conversion from Twitter ads to visits to your page or other specific actions - Here's a vid explaining it https://business.twitter.com/en/help/campaign-measurement-and-analytics/conversion-tracking-for-websites.html

Please could you help set this up? For now we could try adding this to the DI site but in future we may want to add it to the GNR page too. Thanks.

edwinmp:
@Tim-Molyneux this will require access to the DI twitter account - you can go through the Twitter steps and share the code snippet. With that, we'll only need to update Google Tag Manager.

edwinmp:
@davidebukali reopen if there's anything else still pending on this.

PR #1154, opened 2022-06-16
Add twitter website tag

davidebukali:
#1151

### DIFF
--- a/di_website/templates/layouts/default.html
+++ b/di_website/templates/layouts/default.html
@@ -10,6 +10,16 @@
   'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
   })(window,document,'script','dataLayer','GTM-NH32T6V');</script>
   <!-- End Google Tag Manager -->
+  <!-- Twitter universal website tag code -->
+  <script>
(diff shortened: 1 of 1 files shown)


di_website/templates/layouts/default.html
```diff
@@ -10,6 +10,16 @@
...
+  <!-- Twitter universal website tag code -->
+  <script>
+  !function(e,t,n,s,u,a){e.twq||(s=e.twq=function(){s.exe?s.exe.apply(s,arguments):s.queue.push(arguments);
+  },s.version='1.1',s.queue=[],u=t.createElement(n),u.async=!0,u.src='//static.ads-twitter.com/uwt.js',
+  a=t.getElementsByTagName(n)[0],a.parentNode.insertBefore(u,a))}(window,document,'script');
+  // Insert Twitter Pixel ID and Standard Event data below
+  twq('init','o8fck');
+  twq('track','PageView');
+  </script>
+  <!-- End Twitter universal website tag code -->
```
edwinmp, line 22: Why is this required if you're using Google Tag Manager? The whole point of GTM is to avoid this kind of thing ...

davidebukali: Closed this as the tag fires without it included.

Issue #1147, opened 2022-05-23
Related content 


EmmawDI:
**Is your feature request related to a problem? Please describe.**
The Related Content section on our resource/blog pages becomes out of date quickly and needs to be manually updated to direct users to more recent content. As a result, users are unlikely to discover more recent content relevant to their search/interest, instead only being directed to older (and therefore potentially less useful) resources. 

**Describe the solution you'd like**

We have been thinking about how we can ensure that the Related Content on resource/blog pages on the website can be kept relevant and up to date, to make sure that users are directed to our most recent content. 
Would it be possible to change the current Related Links block in the CMS, so that instead of Page being the only option, we can either:
-	Manually select a page, as we can currently, or
-	Choose a country (from the Snippet list), so the site displays the most recent content tagged with that country, or
-	Choose a topic (from the Snippet list), so the site displays the most recent content tagged with that topic
Does this sound feasible?
 

![image](https://user-images.githubusercontent.com/101582869/169798308-0ac0c5c6-91e1-4133-be53-babddc1df7e8.png)

**Describe alternatives you've considered**
The only real alternative to achieve the same result is to manually update each page in the CMS.

pauline-n:
<img width="423" alt="Screenshot 2022-06-30 at 14 05 27" src="https://user-images.githubusercontent.com/71310090/176662385-b060016a-8ecc-404b-ba98-15dd0a15941a.png">
Something like this? @EmmawDI

EmmawDI:
Hi @davidebukali - thank you - I can't seem to log into the staging site currently, is this something you can help with so I can have a look?

davidebukali:
@EmmawDI sure thing, which email do you prefer to use ...

EmmawDI:
emmaw@devinit.org please!

davidebukali:
Sent you mail with details

SimonMurphyDI:
Hello @davidebukali Just used this and it's working well - there's one small change I suggest. At the moment the content is ordered older to newer (left to right) - can we swap this so the newest bit of the content is on the left hand side, so it's the first thing users see?

Thanks!

![image](https://user-images.githubusercontent.com/22001094/204751975-93d9b148-15a0-41d6-857d-4ebc688430ca.png)

davidebukali:
@SimonMurphyDI I've created an issue [here](https://github.com/devinit/DIwebsite-redesign/issues/1269)

SimonMurphyDI:
[...]

PR #1174, opened 2022-07-01
Feature | Related Content Additional Options

pauline-n:
#1147


edwinmp (changes requested):
Add this to every publication page type that has related links ... there's several publication page types besides `PublicationPage`

di_website/publications/models.py
```diff
@@ -486,16 +485,26 @@ def save(self, *args, **kwargs):
...
+            queryset = PublicationPage.objects.filter(topics__in=self.topics)
+            context['related_pages'] = get_related_pages(
+                self, self.publication_related_links.all(), queryset)
+        elif related_options == 'Country':
+            queryset = PublicationPage.objects.filter(page_countries__in=self.page_countries)
+            context['related_pages'] = get_related_pages(
+                self, self.publication_related_links.all(), queryset)
+        else:
+            context['related_pages'] = get_related_pages(
+                self, self.publication_related_links.all(), PublicationPage.objects)
```
edwinmp, line 500: You have this almost exactly repeated in like 5 different places ... Use a function and add it in the `RelatedLinksMixin`

```
def get_related_pages(self, option='Manual', PageClass):
     ....
```

Usage

`self.get_related_pages(related_options, PublicationPage)` - be sure to replace PublicationPage with all the other page types as required.

edwinmp (changes requested):

di_website/publications/models.py
pauline-n: Great, thanks

edwinmp (changes requested):
The country option doesn't seem to work - the links shown aren't for the target country.

Also, got an error when trying to view some of the publications.

image

edwinmp (changes requested):
The country option doesn't seem to work - the links shown aren't for the target country.

Also, got an error when trying to view some of the publications.

image

edwinmp (approved):
No idea if this works, so may as well get it to staging and find out :)

Issue #1159, opened 2022-06-22
Enhance or make clearer how the charts should be used


edwinmp:
e.g. It’s not clear enough that a user can select & compare multiple donors

This seems like a UX issue.

pauline-n:
@edwinmp Below are the design suggestions for enhancing the charts;

1. Moving the "How to use this chart" accordion from below the chart to above it, just before the Select Donors dropdown. 
2. Change 'Select donors' to 'Select up to 2 donors'. 
3. Using a multi-select and autocomplete instead of the dropdown element for the select donors field.

pauline-n:
![GHA Charts](https://user-images.githubusercontent.com/71310090/176855106-401f33ca-a97e-4195-8d65-c6673ddb9cb1.png)

Issue #1169, opened 2022-06-30
Configure the position of the 'How to use this chart' accordion


edwinmp:
Can either be top or bottom of the chart. Refer to ticket #1159 on top design

PR #1176, opened 2022-07-05
Move accordion above graphs

wakibi:
Change position of accordion as per #1159


edwinmp (changes requested):
This wasn't the requirement ... you also tagged the wrong ticket. [This](https://github.com/devinit/DIwebsite-redesign/issues/1169) is the right ticket. As we discussed in the sprint planning, the user must be given an option of having the accordion either at the top or bottom, with the bottom position as the default so that existing charts are not affected.

PR #1204, opened 2022-08-23
 GHA Map Icons

ThatcherK:


edwinmp (changes requested):
Please put them neatly in their own folders, as all the other static files are ... for example, in the patterns library, they're nicely placed in `svg/icons` folders. Also, lowercase naming all through.

edwinmp (approved):

Issue #1212, opened 2022-09-12
Add canonical tags for alias pages


edwinmp:
https://developers.google.com/search/docs/advanced/crawling/consolidate-duplicate-urls

PR #1214, opened 2022-09-14
Feature | Canonical Tags

ThatcherK:
Ticket #1212


ThatcherK:
@edwinmp , the canonical tags have been added, but we still have an unresolved bug with copying already existing pages, the same error appears on the dev site
Screenshot 2022-09-15 at 12 51 15

edwinmp (approved):

Issue #1223, opened 2022-09-20
Copying pages is broken


edwinmp:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[x] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
Another bug originating from the related content feature update in ticket #1147.
Details of the error extract from the email are below:

```
Internal Server Error: /admin/pages/2673/copy/

ValidationError at /admin/pages/2673/copy/
{'related_option_handler': ["Value 'Manual' is not a valid choice."]}

```
**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.
1. Go to any page on the Wagtail CMS
2. Attempt to make a copy of it
3. A 500 error will occur

**Expected Behaviour**
Should be able to copy pages without any errors

Requested Delivery Date: ASAP

Issue #1221, opened 2022-09-20
Common function `get_related_pages` called without a valid queryset causes live page to throw 500 error


akmiller01:
## I'm submitting a ...
Check one of the following options with "x" and add the appropriate label to the issue as well
<pre><code>
[x] Bug report 
[ ] Regression (behaviour that used to work and stopped working in a new release)
</code></pre>

**Describe the Issue**
Line 50 of the project page model.py calls `get_related_pages` without a queryset argument: https://github.com/devinit/DIwebsite-redesign/blob/develop/di_website/project/models.py#L50

This causes `get_related_pages` to jump down to this line and attempt to call `.live()` on a None: https://github.com/devinit/DIwebsite-redesign/blob/1d0871995f975b9993ddddfe27108df24370cc06/di_website/common/base.py#L86

**To Reproduce**
Replace the content below with the steps to reproduce the behaviour.
1. Go here: https://www.devinit.org/admin/pages/627/edit/
2. Delete manually added Related content
3. Click "Preview"
4. See 500 error

Requested Delivery Date: ASAP

edwinmp:
Ah, forgot the fix for this is in another PR ... let me merge it in

edwinmp:
hmm ... Actually, may not be related ... let me verify

akmiller01:
Okay, I've temporarily added 3 related pages for the 3 project pages that were throwing the 500 errors, so you'll need to temporarily delete them to test.

edwinmp:
Yup, not related ... that's two issues tied to this related content change ... I'll take care of both tomorrow.

edwinmp:
This is fixed, thanks @davidebukali.

PR #1224, opened 2022-09-21
Fix | Copy Pages

davidebukali:
#1223 and #1221


edwinmp (approved):
Not sure if this fixes the bug on copying pages as it seems to have nothing to do with it, but as long as it fixes something, lets merge it in.

PR #1226, opened 2022-09-22
Fix | Copying Pages

davidebukali:
#1223


edwinmp (changes requested):
Why are the options uppercase to begin with? Couldn't changing them to lowercase fix it? Seems overkill to create a hook simply to fix a wonky coding choice.

davidebukali:
@edwinmp changing to lower case requires a migration which may affect existing pages and also need a data migration to update. This seemed a shorter route. If few pages have used the feature, I think lowercasing would be ideal.

edwinmp:
> @edwinmp changing to lower case requires a migration which may affect existing pages and also need a data migration to update. This seemed a shorter route. If few pages have used the feature, I think lowercasing would be ideal.

This just went live last week, by accident ... I'm sure we can afford a cleaner solution with no disruption. I'm absolutely certain that the other options (not manual) haven't been used. The original ticket is still in review and hasn't been commented on.

edwinmp:
No data migration needed.

di_website/publications/mixins.py
```diff
@@ -14,9 +14,9 @@
 from .utils import WagtailImageField, get_downloads
 
 RELATED_CHOICES = (
-    ('MANUAL', 'Manual'),
-    ('COUNTRY', 'Country'),
-    ('TOPIC', 'Topic')
+    ('Manual', 'Manual'),
+    ('Country', 'Country'),
+    ('Topic', 'Topic')
```
edwinmp, line 19: This still has upper case letters in it.

edwinmp (changes requested):

edwinmp (approved):

Issue #1199, opened 2022-08-15
Adding link functionality to endnotes 


EmmawDI:
**Is your feature request related to a problem? Please describe.**
Plain URLs are not ideal in terms of their accessibility, as link text should always be unique and descriptive, so having lists beginning "https://www..." gives a difficult user experience. If we could apply links to text in the endnote box, as in the normal WYSIWYG editor, we could hyperlink the name of the source instead of leaving the URL as-is.

**Describe the solution you'd like**
To be able to format endnote text like paragraph text in the WYSIWYG editor. Rather than plain text, to have the option to apply hyperlinks to the text.

This relates to 2.4.9 of the WCAG 2.1 accessibility criteria. 

**Requested Delivery Date:** 31/Oct/2022

edwinmp:
@EmmawDI could you please share a link to a page where you want this changed made.

EmmawDI:
Any resource page type - so factsheet, report, briefing etc. But this page would be a good test - https://devinit.org/resources/oda-2020-2021-key-trends-before-during-emerging-crises/

edwinmp:
**Implementation Details**

- Under `di_website/footnotes/forms.py`, import `from wagtail.admin.rich_text import DraftailRichTextArea`
- Replace the widget on the text form field from `forms.TextArea` to `DraftailRichTextArea`
- Edit `di_website/templates/publications/footnotes.html` to handle richtext content
- You'll need to change the parent component from a paragraph to a div or the footnotes won't render correctly.

edwinmp:
@EmmawDI this is deployed for testing on [dev](https://dev.devinit.org/admin/). We couldn't add a WYSIWYG editor to it, so the only way to allow richtext features for this is to use Markdown syntax.
So for the links you want to change, e.g. https://google.com - you'll need to rewrite that as `[Google](https://google.com)`.
i.e The text you want to show on the footnotes is in the square brackets while the actual link is in parentheses. A better explanation can be found [here](https://www.markdownguide.org/basic-syntax#links) as well as the full list of Markdown syntax.

EmmawDI:
Apologies I had not had a chance to look at this before - I think the principle of this is good, but on the Dev site it doesn't seem to be working for me. I've tested it here https://dev.devinit.org/admin/pages/2245/edit/

edwinmp:
[...]

PR #1233, opened 2022-10-03
Feature | Richtext Footnotes

edwinmp:
Ticket #1199 

There's just one problem left ... we can't `urllize` the footnotes and have them as Richtext at the same time. So this change will have an effect on existing footnotes i.e the links won't appear as links anymore. So, not ideal - I'm thinking of adding another field (`use_richtext`) to control the rendering - naturally, this would default to false so existing footnotes wouldn't be affected. It's a more complicated change than it looks though.


edwinmp:
> There's just one problem left ... we can't `urllize` the footnotes and have them as Richtext at the same time. So this change will have an effect on existing footnotes i.e the links won't appear as links anymore. So, not ideal - I'm thinking of adding another field (`use_richtext`) to control the rendering - naturally, this would default to false so existing footnotes wouldn't be affected. It's a more complicated change than it looks though.

Sorted it in a kinda dirty way - should wagtail change their implementation down the road, legacy footnotes could break.

akmiller01 (approved):
Tested locally. This appears to be working well. Does `npm run build` run as a part of the deployment process? Or do you need to commit the build to this PR @edwinmp ?

edwinmp:
> Tested locally. This appears to be working well. Does `npm run build` run as a part of the deployment process? Or do you need to commit the build to this PR @edwinmp ?

It's part of the deployment process.

Issue #1181, opened 2022-07-11
Adding accordion feature to the DI site


SimonMurphyDI:
**Is your feature request related to a problem? Please describe.**
I love the accordion feature on the GNR site - you can see it in action here: https://globalnutritionreport.org/resources/naf/faqs/

**Describe the solution you'd like**
We'd like the capacity to use this on the DI site, on the following page types
- audio and visual media page (priority)
- short publication page
- the child pages of the publications page
It might be we need it on other pages, but I think it's worth adding it as we find we need it

It would be good if these can use the DI website font and colours (nb the feature might need to be designed in six colours so it can work on the resource pages)

**Requested Delivery Date:** Ideally we'd like to use this to host some long transcripts of podcasts - if it were possible to do over the next few weeks that would be wonderful, but I understand you might have other priorities

edwinmp:
@SimonMurphyDI we've split the design bit into its own ticket #1183

SimonMurphyDI:
Thanks Edwin

Happy for you to do whichever is easiest - either using the GNR or DI version. 

Whichever you build

edwinmp:
@wakibi ^^

SimonMurphyDI:
@wakibi - I also wondered if there is any "cut-off" for the total number of words that can be held in an accordion? (Would be handy to know for DI site and also GNR site
Thanks!

edwinmp:
@wakibi this has been active forever. If there's a branch with some work on it, please push. I'll have someone else complete it so u can focus on IATI.

SimonMurphyDI:
Hello @ThatcherK  - we've been asked for an ETA for this - please can you provide? Thanks!

edwinmp:
@SimonMurphyDI this is done and should be available for testing on staging ...

SimonMurphyDI:
Great thanks!

SimonMurphyDI:
Hi Edwin - thanks for this. We'd like the WYSIWYG capacity for accordions - sharing the GNR back end, which works really well. 
![image](https://user-images.githubusercontent.com/22001094/195090417-1b81e0e4-ac54-41c7-be03-fc061ae5e8ec.png)

edwinmp:
@SimonMurphyDI the implementation on dev does have WYSIWYG capacity ... more Wagtail 4 changes on how that works.

[...]

PR #1234, opened 2022-10-04
Feature|Accordion

ThatcherK:
Ticket #1181


di_website/common/blocks.py
```diff
@@ -3,28 +3,49 @@
...
+GREEN = 'leaf'
+COLOUR_CHOICES = (
+    (RED, 'Red'),
+    (BLUE, 'Blue'),
+    (PINK, 'Pink'),
+    (YELLOW, 'Yellow'),
+    (ORANGE, 'Orange'),
+    (PURPLE, 'Purple'),
+    (GREEN, 'Green')
+)
```
edwinmp, line 34: This is redundant as it already exists (almost identically) in the publications modal page. This make sense since this kind of theming is only applicable to publication (resource) pages, and not all of the e.g AudioVisual don't support it. See my next comment for how to handle theming for these different types of pages.

di_website/common/blocks.py
```diff
@@ -3,28 +3,49 @@
...
+class AccordionBlock(StructBlock):
+    colour = ChoiceBlock(choices=COLOUR_CHOICES, default=RED)
```
edwinmp, line 36: This isn't the way to do it ... colours (themes) are tied to the entire page, and pages that support this kind of theming already have this colour field. So your accordion CSS should be based off those individual body colours e.g pages with the `rose` theme have the class `body--rose` added to the body element. So this is redundant as well.

image

di_website/publications/static/publications/css/style.css
```diff
@@ -0,0 +1,31 @@
...
+  background-color: #EB632B;
+}
+
+.accordion-lavendar {
+  background-color: #89328F;
+}
+
+.accordion-spacing .accordion__list-item:first-child {
+  margin-top: 0;
+}
```
edwinmp, line 31: You'll have to change these to be dependent on the body's thematic classes i.e.
`.body--rose .accordion__heading {...}`

di_website/templates/blocks/accordion_block.html
```diff
@@ -0,0 +1,17 @@
+{% load wagtailcore_tags string_utils %}
+
+<div class="max-meter max-meter--center accordion-spacing">
+  {% for item in self.accordion_items %}
+    <li class="accordion__list-item">
+      <button type="button" class="accordion__heading accordion-{{self.colour}}">
```
edwinmp, line 6: Won't need specific colour classes.

di_website/templates/includes/scaffold/header-assets.html
```diff
@@ -10,3 +10,4 @@
 <script src="{% static 'js/libs/modernizr.js' %}"></script>
 {% include 'includes/scaffold/favicon.html' %}
 <link href="{% static 'css/di_website.min.css' %}" rel="stylesheet" />
+<link href="{% static 'publications/css/style.css' %}" rel="stylesheet" />
```
edwinmp, line 13: Nope ... this is shared across the entire app, so why would we need publications CSS for non-publications pages? Unless u added the classes in the wrong place. In which case, add them to the global CSS file `di_website.css` ... you'll need to minify it and update the minified version of the file ... this isn't tied to a task runner so u'll need an online minifier ... shouldn't be a problem.

edwinmp (changes requested):

ThatcherK:
@edwinmp , changes made

di_website/static/css/di_website.css
```diff
@@ -191,3 +191,43 @@ select optgroup {
...
+
+.body--marigold .accordion__heading {
+  background-color: #EB632B;
+}
+
+.body--lavendar .accordion__heading {
+  background-color: #89328F;
+}
+
+.accordion-spacing .accordion__list-item:first-child {
```
edwinmp, line 223: In addition to this, give the last child `margin-bottom: 2em` ... otherwise it looks cramped in some non-section usages. The one below is for the AudioVisual page type. Perhaps only add this margin for when the accordion doesn't have it's own section ... I think the distinguishing class for a bottom margin is `wysiwyg`.

`.wysiwyg .accordion-spacing .accordion__list-item:last-child` - maybe

image

edwinmp (changes requested):
Add this to the section StreamField ... it's not there. This is for AudioVisual pages and
image

edwinmp (approved):

PR #1236, opened 2022-10-10
Fix | Footnotes

edwinmp:
Ticket #1199 

Our original footnotes implementation doesn't work when run through docker because for some reason, the Draftail Editor doesn't like add-ons that use it - it's a self-reference thing. Anyway, I thought we may need to abandon it completely, calling it a Time vs Value thing, but then I remembered Markdown. Anyway, the requirements of the ticket can be met by using markdown syntax i.e. for a link `[Link](#)` should be turned to HTML as a link.
The `markdownify` filter preserves existing behaviour since it also does what `urlize` does.

Also removed some redundant CSS for the Wagtail 4 upgrade and added an exception handler for a `wagtail-metadata` error when creating image renditions. Lastly, the help icon was missing on help blocks since the Wagtail 4 upgrade ... fixed that too.

### DIFF
--- a/di_website/common/mixins.py
+++ b/di_website/common/mixins.py
@@ -1,7 +1,9 @@
+from django.conf import settings
 from django.db import models
 
 from wagtail.models import Orderable
 from wagtail.fields import RichTextField, StreamField
+from wagtail.images.models import SourceImageIOError
 
 from di_website.common.blocks import BaseStreamBlock, SectionStreamBlock, TypesetStreamBlock, TypesetFootnoteStreamBlock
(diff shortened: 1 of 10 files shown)


di_website/common/mixins.py
```diff
@@ -30,6 +32,12 @@ def get_meta_title(self):
     def get_meta_twitter_card_type(self):
         return 'summary_large_image'
 
+    def get_meta_image_rendition(self):
+        try:
+            return super(CustomMetadataPageMixin, self).get_meta_image_rendition()
+        except SourceImageIOError:
```
ThatcherK, line 38: Nice catch, I was getting these errors alot

ThatcherK (approved):

PR #1256, opened 2022-11-07
Only use production vars for master branch or non-beta release

akmiller01:


.github/workflows/deploy_CI.yml
```diff
@@ -34,18 +34,18 @@ jobs:
...
-          echo "SITE_URL=dev.devinit.org" >> $GITHUB_ENV
-          echo "DATA_SITE_URL=dev.devinit.org" >> $GITHUB_ENV
+          echo "WWW_SITE_URL=www.devinit.org" >> $GITHUB_ENV
+          echo "SITE_URL=devinit.org" >> $GITHUB_ENV
+          echo "DATA_SITE_URL=data.devinit.org" >> $GITHUB_ENV
 
-      - name: Set env for production
-        if: endsWith(github.ref, '/master') || contains(env.TAG, '-beta') != true
+      - name: Set env for production if release and not beta
+        if: startsWith(github.ref, 'refs/tags/v') && ( contains(env.TAG, '-beta') != true )
```
wakibi, line 48: I think the checks on line 38 should be the same as on this line... 
Also propose to remove line 7, so we only deploy to production by way of tags strictly, and not pushing to master. That should go a long way in reducing the confusion (or complexity) of this

wakibi (changes requested):

.github/workflows/deploy_CI.yml
akmiller01: Okay, I removed line 7 and removed the whole deploy to live on master branch ref section.

wakibi (approved):
LGTM

PR #1282, opened 2022-12-13
Blog footnote markdown

akmiller01:


di_website/templates/blog/footnotes.html
```diff
@@ -1,10 +1,10 @@
-{% load string_utils wagtailcore_tags %}
+{% load string_utils wagtailcore_tags markdownify %}
 {% if page.footnotes_list %}
     <div class="m-notes">
       <h3 class="m-notes__title">Notes</h3>
       <ul class="m-notes__list">
         {% for item in page.footnotes_list %}
-          <li><sup><a id="{{ item.footnote_id }}" href="#{{ item.source_id }}">{{ forloop.counter }}</a></sup>{{ item.text | urlize }}</li>
+          <li><sup><a id="{{ item.footnote_id }}" href="#{{ item.source_id }}">{{ forloop.counter }}</a></sup>{{ item.text | urlize | markdownify }}</li>
```
edwinmp, line 7: If I remember correctly, markdownify on its own should be able to urlize as well.

akmiller01: Just gave it a test, you're correct. Pushed a commit with just markdownify.

edwinmp (approved):

Issue #1285, opened 2023-01-03
Enhance accessibility on the Short Publication Page


edwinmp:
Perform an accessibility audit on the Short Publications Page and fix any flagged accessibility issues.

Resources: [WAVE](https://chrome.google.com/webstore/detail/wave-evaluation-tool/jbbplnpkjmmeebjpijfedlgcdilocofh)

For full coverage, you'll need to create a test page that makes use of every possible StreamField item.

edwinmp:
Hi @EmmawDI ... We've made accessibility enhancements for the Short Publications page and this has been deployed to [staging](https://dev.devinit.org/). Please review when u have a moment.

edwinmp:
> Do you have a record of what changes were made?

@davidebukali ^^


> It seems as though text contrast is a recurring problem, especially with linked text. Do you think this is something we need to look to update across the site?

Yes, some of our colours have been flagged as not being of great contrast, not just on text. To ensure uniformity, it's not a change we should make in bits if it's happening all over the site.

EmmawDI:
Sure, that makes sense. Simon is going to talk to Anna about updating some colours across the site so will get back to you asap on that

davidebukali:
Yes we have a record of the changes which were mostly to the css. A few updates were made to the structure of html. These could apply to other publication types since they share sections.

EmmawDI:
Anna is happy for us to make colour updates necessary to meet accessibility requirements. 
- Hyperlinked text (including "share" links and notes) and pull-out quote text: if this is changes to Colour 5 of the respective hue, we should be meeting the [WCAG 1.4.6 AAA criteria for enhanced contrast at 7:1. ](https://www.w3.org/TR/WCAG21/#contrast-enhanced)
- More decorative elements (ie white text on coloured boxes for CTAs, certain buttons, table headings etc) will also need to be made darker, changing to colour 5 across the board. Note the default colour of the pages doesn't need to change from Colour 1 as these are brand colours and do not have text directly on them. 

Would it be possible for this change to be made in staging, for us to review and present to Anna? As you mentioned, these changes will affect the whole site so we want to avoid any situation where one page type looks noticeably different from another. 

On a broader note - what is the plan going forward to continue the audit? Simon has flagged the [pattern library](https://development-initiatives.surge.sh/page-templates/) to me, which we could use as a basis to record issues found and fixes implemented, so we can see crossover between page types. 

[...]

PR #1287, opened 2023-01-09
Feature | Short Publication Accessibility

davidebukali:
Ticket #1285


ThatcherK (changes requested):
There are still some alerts and errors I see that could be fixed

1. In the image below, the description text in description lists is small

Screenshot 2023-01-09 at 14 54 45
2. In the footer
Screenshot 2023-01-09 at 18 21 25

3. Am also getting this alert for images in the case study infographic as well as the stand alone infographic image
Screenshot 2023-01-09 at 18 28 55

4. I have also noted that when orange, blue, yellow, and green are used as a theme colour, we have multiple contrast errors. I wonder what decision will be taken on that?

ThatcherK:
Good job with the contrast errors! 
Just one alert I missed here, that can easily be resolved
Screenshot 2023-01-18 at 11 16 50

edwinmp:
@davidebukali I hope this has covered every existing `StreamField` type.

ThatcherK (approved):

Issue #1295, opened 2023-02-21
Update to Focus Area page type


EmmawDI:
We’re looking to change the contents of [this page](https://devinit.org/what-we-do/what-we-are-working-on/?nav=header) and its child pages. It would be great if you could provide us with an estimate for how long this work will take and whether it will be possible to **complete it before 3 March** (we would need time to upload content before the planned launch of one of these pages on International Women’s Day, 8 March). 
What we’d need: Focus Area page type changes
-	Our plan for the revised structure would involve Focus Area pages at two levels of site navigation – the current “what we are working on” page, and as child pages of that page. 
-	Currently, within each Focus Area, it’s only possible to link to Project page types in the Projects section. We would like this to be changed so that we can link to any page type, including resources. 
-	Can we apply the same logic as the [related content feature](https://github.com/devinit/DIwebsite-redesign/issues/1147)? In the child Focus Area pages, we would like to have the option for the “Project” sections to automatically populate with resource pages tagged with a specific topic. 
Happy to discuss this if it’s easier.

edwinmp:
> * Our plan for the revised structure would involve Focus Area pages at two levels of site navigation – the current “what we are working on” page, and as child pages of that page.

So you need the `Focus Area Page` type to allow children of its own type?
This is an easy change.


> * Currently, within each Focus Area, it’s only possible to link to Project page types in the Projects section. We would like this to be changed so that we can link to any page type, including resources.

Pages added to this section are titled `Key Projects`. I recommend that we make this configurable in the CMS since this will now be linking to more than just projects.
Both are easy updates.


> * Can we apply the same logic as the [related content feature](https://github.com/devinit/DIwebsite-redesign/issues/1147)? In the child Focus Area pages, we would like to have the option for the “Project” sections to automatically populate with resource pages tagged with a specific topic.

This will be tougher. These pages were never meant to connect in this way. Still, we'll manage, though it'll need a great deal of testing. For clarity, you only want the two options i.e. `manual` & `topic`? Or do we replace the `manual` option with the `topic` option.


Overall, the changes above shouldn't take longer than 3 days (including reviews & testing).

EmmawDI:
Hi Edwin, thank you so much for your response. 

1 - yes, exactly. 
2 - making the title configurable sounds ideal, thank you
3 - Ideally we would have manual and topic options.  If it would be helpful for someone on our team to help with testing, do let us know and I'm sure we can plan some time. 
 
[...]

PR #1297, opened 2023-03-02
Feature | Update Focus Area Page Type

davidebukali:
Ticket #1295


di_website/common/constants.py
```diff
@@ -28,6 +28,7 @@
     'anchor',
     'footnote'
 ]
+ALT_MAX_RELATED_LINKS = 2
```
edwinmp, line 31: If this is only going to be used for a specific page type, add it somewhere close to wherever it's being used. It doesn't look useful here.

di_website/project/models.py
```diff
@@ -135,14 +160,45 @@ class FocusAreasPageLink(Orderable):
         verbose_name='Image',
         help_text='Add an image to this focus area'
     )
+    related_page_section_title = models.CharField(blank=True, max_length=255, default='Key Projects and Publications', verbose_name=SECTION_TITLE)
```
edwinmp, line 163: `Key projects and publications`

edwinmp (changes requested):

edwinmp (approved):

Issue #1306, opened 2023-05-04
Show most recent related content in publications


davidebukali:
**Is your feature request related to a problem? Please describe.**
After adding related content to a publication based on country or tag, the publication does not display the most recent related content.

**Describe the solution you'd like**
We would like the most recent related content to always be displayed for all available related options including by country, tag or manually selected

**Project**
This is connected to the DI Website.

SimonMurphyDI:
Thanks David, I applied it to this page to check it. 
https://dev.devinit.org/admin/pages/1281/edit/
The page is topic-tagged as humanitarian
But the automatic results shows content from 2021 - which isn't the latest humanitarian work. Is there a bug to check here?
Best
Simon
![image](https://github.com/devinit/DIwebsite-redesign/assets/22001094/230c71a1-bdbc-485c-9be9-69a8721d9bee)

davidebukali:
Hi @SimonMurphyDI I just searched [resources](https://dev.devinit.org/resources/?types=&topic=humanitarian-assistance_1&country=&q=) page for humanitarian assistance topic and it shows similar results. Maybe the latest content has not been copied over to staging from the live website.

SimonMurphyDI:
Ah my bad - there is more recent humanitarian content it's just not tagged. Looks like this is good to go!

edwinmp:
@davidebukali not sure if this has been deployed. If it has, please close.

PR #1307, opened 2023-05-07
Sort related content in publications

davidebukali:
Ticket #1306


wakibi (changes requested):
@davidebukali I see these are over 3 lines of changes repeated across two different sections. Maybe it's time to actually create a new function to do this? DRY!

wakibi (approved):

Issue #1261, opened 2022-11-17
Reducing space between accordions/anchor links and surrounding text


LCowie:
In the current page layout, there are large spaces between accordion blocks and surrounding text. This also applies to anchor links - there are large spaces between anchor links and surrounding text too.

Please can you reduce the size of the spacing between accordions/anchor links and the text that precedes or follows them? This will make outputs appear more compact visually. 

Requested Delivery Date: as soon as possible or 24/11/2022

k8hughes:
@LCowie do you know what budget code we should be charging this work to?

SimonMurphyDI:
@k8hughes 
Sorry - we don't have a code yet. Have you and Anna heard back from Anthony? If there's no code then I guess we have to pause website improvements like this, because we won't know where to timesheet the work. 
If that's the case, I can go through all the tickets we've raised and ask them to pause
Best
Simon

k8hughes:
I haven't had any communication about budgets so I will chase Anthony.  I was hoping comms colleagues had heard and I'd just missed something!  Don't pause it all yet, wait till I speak to Anthony.

edwinmp:
@LCowie not yet. A bot had closed the ticket. I've pinned it so the bot won't close it again.

LCowie:
Hi Edwin,

Thanks for the update, it would be great to get this issue fixed so that we can use accordions and anchor links more effectively in our outputs.

All the best,

Lucy

Lucy Cowie | Content & Publications Officer

I work Mondays, Tuesdays and Thursdays
Development Initiatives, First Floor Centre, The Quorum, Bond Street South, Bristol, BS1 3AE, UK
T: +44 (0) 1179 272 505

At DI we value and respect flexible work arrangements. I choose to work flexibly and sometimes send emails outside normal working hours – no need to respond outside yours.
***@***.***

Read more about our work and its impact<https://devinit.org/what-we-do>
Development Initiatives is the trading name for: Development Initiatives Poverty Research Ltd, a not-for-profit company. Company No. 06368740 and DI International Ltd. Company No. 5802543, both registered in the UK.  Registered Address: First Floor Centre, The Quorum, Bond Street South, Bristol, BS1 3AE, UK.

davidebukali:
@LCowie kindly include a screenshot or link with a page affected by this issue; Are the accordions showing `Our Expertise` on [this](https://devinit.org/what-we-do/?nav=header) page affected ?

LCowie:
[...]

PR #1308, opened 2023-05-15
Feature | Accordion Spacing

davidebukali:
Ticket #1261


edwinmp (changes requested):
Seen what you've done with this & have some thoughts.
You've created 2 CSS classes and are applying the same styles to them. You're then using these same classes on publication section block, using an if statement to decide which class to apply. The problem is, both classes add the same style. So why the 2 classes?

For such situations, create a CSS class with a generic name related to its intent - in this case, something to do with `margins`. On your sections, your if statement then checks for either `accordion` OR `rich_text` and adds that one class.

edwinmp (approved):

Issue #1310, opened 2023-05-22
Bug on preview of publication pages when Published date field is left blank


ThatcherK:
## I'm submitting a ...
Bug report

**Describe the Issue**
This error is observed on preview of a short publication when the  Published date field which is an un-required field is not filled
`'Page' object has no attribute 'published_date'`

**To Reproduce**
1. Go to resources in the wagtail admin
2. Add new child page of type short publication
3. Start editing the page with the preview window open

This was first observed on staging, so the bug is definitely on the live site as well

PR #1311, opened 2023-05-22
Fix | Published Date Bug

ThatcherK:
Ticket #1310

### DIFF
--- a/di_website/publications/mixins.py
+++ b/di_website/publications/mixins.py
@@ -89,11 +89,11 @@ class PublishedDateMixin(models.Model):
 
     @cached_property
     def publication_date(self):
-        if self.published_date:
+        if hasattr(self, 'published_date') and self.published_date:
             return self.published_date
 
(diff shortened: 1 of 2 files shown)


di_website/publications/mixins.py
```diff
@@ -89,11 +89,11 @@ class Meta:
 
     @cached_property
     def publication_date(self):
-        if self.published_date:
+        if hasattr(self, 'published_date') and self.published_date:
```
edwinmp, line 92: This page will always have a `published_date` property, so this isn't required. This error was originating from the parent on line 96. Still, not harming anyone.

edwinmp (approved):

edwinmp (approved):

edwinmp:
If there's an easily fixable problem that's breaking the release, please fix it in here - no need for a separate ticket for it.

edwinmp (approved):

PR #1342, opened 2023-08-04
Replace edge ngram tokenizer with ngram to prevent partial matches

akmiller01:
Right now a search for "disability" matches "disbursement" because they both begin with "dis". https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-edgengram-tokenizer.html

```
For example, if the max_gram is 3 and search terms are truncated to three characters, the search term apple is shortened to app. This means searches for apple return any indexed terms matching app, such as apply, approximate and apple.
```

So I think we should swap this for standard tokenizer, to make behavior much more predictable: https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-tokenizer.html

### DIFF
--- a/di_website/settings/base.py
+++ b/di_website/settings/base.py
@@ -166,12 +166,6 @@ AUTH_PASSWORD_VALIDATORS = [
 
 if config('ELASTIC_SEARCH_URL', ''):
 
-    elastic_search_tokens = [
-        'letter',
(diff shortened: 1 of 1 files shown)


davidebukali (approved):
LGTM!

wakibi (approved):

PR #1346, opened 2023-08-28
Refactor to simplify/update tech stack

akmiller01:


ThatcherK:
@akmiller01 great job here! Looks like the site will finally stop being slow. One question, why are we moving away from elastic search?

akmiller01:
> @akmiller01 great job here! Looks like the site will finally stop being slow. One question, why are we moving away from elastic search?

Thanks for reviewing, @ThatcherK ! Issues with ES are actually what first prompted this refactor. The ES index was periodically going missing, which was causing search pages to 500. Chris and I thought it might be a "Meow attack," but we couldn't find the vulnerability, so upgrading everything possible was the best course.

That, and it seems like ES might have been overkill for our needs. This psql based search seems to be just as fast, and reuses the dependency Wagtail needs anyway.

ThatcherK (approved):

Issue #1353, opened 2023-11-14
Homepage - ability to create duplicate sections


GeorgeCarver:
**Is your feature request related to a problem? Please describe.**
We would like additional flexibility on the homepage. It's our understanding that we can't change the order of the sections on the homepage, and we can't have multiple/duplicate sections. 

**Describe the solution you'd like**
We would like to be able to re-order sections of the homepage if we wanted to. We would also like to be able to create duplicate sections. E.g. so if we want to use a 'featured content' section, then have some more sections, then have another 'featured content' section.

**Links to pages**
Homepage: https://devinit.org/ 
Homepage in Wagtail: https://www.devinit.org/admin/pages/3/edit/

**Project**
Growth and external relations --> Website content updates.

**Budget Code**
P0589-2.01.02.03 or P0589-2.01.02.02

**Additional context**
![image](https://github.com/devinit/DIwebsite-redesign/assets/62111796/aa533c06-e4d0-45a7-90e3-40db967b2996) 

**Requested Delivery Date:**
By 24/11/2023 if possible, as this is a priority for COP28.

ThatcherK:
@wakibi , @akmiller01 , this sounds like a totally new home page(especially the content), and using a streamBlock for the content, something similar to what we have in publication pages?

akmiller01:
@ThatcherK I think it would involve a StreamField, but I don't know if it would require a full refactor. I think we could turn many of the existing homepage template sections into Blocks.

A strategy that Fffunction has used for transitioning a home page like this in the past was to not erase the previous fields, add the new StreamField, and also add a Boolean field with a toggle to switch the home page content from the past fields to the new streamfield. That way, streamfield content can be added gradually and previewed, and then when it's all ready, the boolean can be ticked to make the change (instead of the home page disappearing when a migration is made).

ThatcherK:
Sounds like a good strategy, I was worried about the migration!

GeorgeCarver:
Hello all! Thanks for getting onto this so quickly. Is it going ok? Let me know if you need any more information or feedback from me. And do you think this will be do-able in time for this Friday?
Best,
George

ThatcherK:
Hi @GeorgeCarver I have already started work on this, will be putting it up for review, then I will also let you know when to test soon!

GeorgeCarver:
[...]

PR #1355, opened 2023-11-20
Feature | Home Page Content

ThatcherK:
Ticket #1353


ThatcherK:
@wakibi @akmiller01 , this is now live on staging https://dev.devinit.org/admin/pages/3/edit. To test 

1. Go to admin page of the HomePage
2. Below the Sections , there is a new field called New content, this will help with switching to the new content arrangement once editing is done. So it should be checked, inorder to reflect the new content while hiding the old content in the template.
3. The new content streamblock is under the `Content` Panel, This has Featured work, Featured content and Sections

wakibi (approved):

PR #1399, opened 2024-07-16
Base64 encoded memorial

akmiller01:
Thought this might be something nice for any developer that opens their console on our home page @wakibi . Hope you're doing alright.


wakibi (approved):
Good idea!

